PR #76, opened 2022-02-17
feat: direct imports from validation index

### LINKED ISSUE #52: Code completions for imports not working
This is based on feedback from telegram.

Completion support on the import string is currently providing no suggestions. We think this is a bug rather than a missing feature.

image

## Reproduce

Press ctrl space while typing an import path -> It would be great if it gave suggestions of package paths (and their contracts) that I can import from.

Specifically, packages that already exist in my installed npm libraries

kanej:
This PR improves import completions by fixing some small issues that collectively give a better experience and adding a partial implementation of "direct imports" based on the validation index.

Fixes #52

## Preview

imports

## Implementation details

* The `"` and `'` characters now auto-close, this increases the likelihood of parsing detecting the import statement and allowing completions
* The `"` character is now a trigger character for completions, we immediately return no completions (no visual indication to user) if we are not in an import statement
* There is a bug fix for completions on the `.` character to remove the `_;` that is added to improve parsing, this was killing all completions in cases like `import ".`
* the display text and insert text now vary to allow the user to go from `""` to `"."` to `"./"` and see the same options but have the insert alter appropriately
* for the empty import we scan the validation index and find all the files in `node_modules` we then find the base packages for each (e.g. `@openzeppelin`, `@ens` and include those as completion options
* for direct imports completion on slash (e.g. `@openzeppelin/`) we complete for any validation index files that are match that path taking `node_modules` as the root
* the `findSolFiles` function in analyzer has been externalized and then injected in to allow turning it off in tests (so I can more easily control the validation index files)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+  "files.exclude": {
+    "client/out": true,
+    "server/out": true
+  }
+}
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -4,6 +4,7 @@ import { createConnection, ProposedFeatures } from "vscode-languageserver/node";
 import setupServer from "./server";
 import { compilerProcessFactory } from "@services/validation/compilerProcessFactory";
 import { ConnectionLogger } from "@utils/Logger";
+import { WorkspaceFileRetriever } from "@analyzer/WorkspaceFileRetriever";
 
 Sentry.init({
   // Sentry DSN. I guess there's no other choice than keeping it here.
@@ -18,12 +19,13 @@ Sentry.init({
 // Create a connection for the server, using Node's IPC as a transport.
 // Also include all preview / proposed LSP features.
 const connection = createConnection(ProposedFeatures.all);
+const workspaceFileRetriever = new WorkspaceFileRetriever();
 
 const logger = new ConnectionLogger(connection, (err) =>
   Sentry.captureException(err)
 );
 
-setupServer(connection, compilerProcessFactory, logger);
+setupServer(connection, compilerProcessFactory, workspaceFileRetriever, logger);
 
 // Listen on the connection
 connection.listen();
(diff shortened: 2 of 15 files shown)


the-vampiire:
how do i get this to work? the only thing it suggest (cmd+space) are the local words

image

vscode: 1.66.2
solidity + hardhat ext: v0.2.2

the-vampiire:
i think i figured it out. i have a monorepo with solidity and client code:

```sh
project/
  client/
  solidity/ <-- hardhat / contracts etc
```

if i open just the solidity dir in vscode it works. however when i have the entire project opened it does not.

@kanej is there a way to set a path for the intellisense? so instead of looking top level it can look at "hardhat-project-root"?

the-vampiire:
some recordings to show what i mean (this is with a different project)

```sh
project/
  hardhat/ <-- contracts
  frontend/
  ...
```

in both cases i waited for hardhat vscode to index the files

> opening entire project (path: `hardhat/contracts/TodoList.sol`)

does not provide intellisense

https://user-images.githubusercontent.com/25523682/164933845-d6afdfa6-f05b-4d73-87fc-3e2397fb3515.mov

> opening hardhat only `contracts/TodoList.sol`

provides intellisense


https://user-images.githubusercontent.com/25523682/164933903-7a717b7a-7d8e-4fe8-8b51-e481bc93db20.mov

kanej:
Hey @the-vampiire, thanks for the videos, they always make it far easier to understand an issue.

You are correct. Currently to get import completions you need to open a vscode workspace folder on the folder with the `hardhat.config.{js,ts}` in it. So in a monorepo you would need to open up the package folder. This limitation is captured as an issue here: https://github.com/NomicFoundation/hardhat-vscode/issues/138.

Better monorepo support is being worked on here: https://github.com/NomicFoundation/hardhat-vscode/pull/160, and is intended to make the next release. Specifically you should be able to open up your vscode in the root of the monorepo and get import completion support and inline validation support in your packages subfolders.

PR #120, opened 2022-03-15
feat(vscode): extenstionKind

sambacha:
extensionKind is a property in the extension manifest. It allows extensions to specify a preferred running location. That can be the machine that has the workspace (workspace) or the user interface (ui). If an extension can run on both, it can specify an order of preference. 

[see https://code.visualstudio.com/api/advanced-topics/extension-host#preferred-extension-location](https://code.visualstudio.com/api/advanced-topics/extension-host#preferred-extension-location)

- [ ] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team.
- [x] I didn't do anything of this.

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,10 @@
     "Linters",
     "Formatters"
   ],
+  "extensionKind": [
+    "ui",
+    "workspace"
+  ],
   "keywords": [
     "solidity",
     "ethereum",


kanej:
@sambacha thanks for this. Looking over the docs you linked I suspected the hardhat vscode plugin always requires the workspace as we need to boot a node-base language server to track the files in the workspace. From the docs, I think that would be:

```json
{
  "extensionKind": ["workspace"]
}
```

Does that makes sense or have I misunderstood the config property?

sambacha:
workspace and ui should both be enabled, with workspace taking precedence per your comments. 

This only effects users who are using vscode.dev, github for the web codespaces, container environments, etc. 


Unless you define the settings for web-extension it shouldn't matter(web extensions dont have nodejs globals like path or fs, have cors enabled, basically a browser mode extension)

sambacha:
Note, for the web codespaces is different than regular codespaces as you have ssh access and proper userid etc etc

PR #164, opened 2022-04-21
docs: add formatting instructions

kanej:
Add instructions for using hardhat vscode for formatting on solidity files.

## Preview

To see the extension.md as rendered markdown: https://github.com/NomicFoundation/hardhat-vscode/blob/docs/add-formatting-instructions/EXTENSION.md

image

## Warning

The links will have to be updated to main before after review.

Linear HHVSC-142


EXTENSION.md
```diff
@@ -22,6 +22,53 @@ Built by the [Nomic Foundation](https://nomic.foundation/). [We’re hiring](htt
 
 Some features (e.g. inline validation) are still experimental and are only enabled within a [Hardhat](https://hardhat.org/) project, this is a limitation that will be lifted with future releases.
 
+## Setup
+
```
alcuadrado, line 26: What about adding something like this?

```suggestion

This extension should work without any configuration. If its formatting functionality isn't working, please read on.
```

kanej: Added, thanks.

PR #198, opened 2022-06-06
feat(validation): cache solc input

### LINKED ISSUE #55: Improve compilation speed
Hardhat leverages caches internally to speed up compilation. Those are turned off when doing compilation for error/warning display in vscode.

We should explore ways of taking advantage of the caches.

kanej:
Within the worker (`workerState`) we retain the last input to solc calculated by the preprocessing steps. If the same uri with the same solc version and imports comes in again, we reuse the same input with an updated content for the one changed file.

This gives a performance improvement (the preprocessing is the larger part of validation in big codebases) for the most common case which is lots of changes in a single file.

Fixes #55

## TODO

* [x] implement caching based on changed doc
* [x] invalidate cache if changes (from file-system) come in on other files
* [x] bundling into vsix

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -17,9 +17,36 @@
 **/node_modules/
 **/src
 **/test
+**/scripts
 
 client/package.json
 server/package.json
 
 **/docs
 !**/docs/images/hardhat-vscode-logo.png
+
+!server/node_modules/@nomicfoundation/solidity-analyzer/package.json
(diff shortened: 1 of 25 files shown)


kanej:
After a round of testing, two bugs have come up:

* caching suppresses import line errors (which happen during preprocessing), made worse by the change text vs text on disk issue
* renames can cause an error line, due I suspect to a race condition between the changes on other pages vs the change on the open text editor hitting the validation thread.

PR #218, opened 2022-08-03
feat:  Support intermediary stars in block comments #75

### LINKED ISSUE #75: Support intermediary stars in block comments
If you open a block comment and hit enter to go to a new line, the `*` should be autoinserted:

Current:
```solidity
  /**
  | <- cursor
   */
  function foo2() public virtual {}
```

Preferred:
```solidity
  /**
   * | <- cursor
   */
  function foo2() public virtual {}
```

This happens in typescript and looks to be coming from the `onEnterRules` in the `language-configuration.json`: https://github.com/microsoft/vscode/blob/888677ed95e58f33e8a9c16d6ee259b1e35e190b/extensions/typescript-basics/language-configuration.json#L124

antico5:
When using multiline comments, intermediate stars (*) are added automatically with correct indentation.

Closes #75

### DIFF
--- a/test/integration/helpers/files.ts
+++ b/test/integration/helpers/files.ts
@@ -18,7 +18,7 @@ export const withRandomFileEditor = async (
     editor: vscode.TextEditor,
     document: vscode.TextDocument
   ) => Promise<void>
-): Promise<boolean> => {
+): Promise<void> => {
   const cursorIndex = text.indexOf(CURSOR);
   const file = await createRandomFile(text.replace(CURSOR, ""), fileExtension);
   const document = await vscode.workspace.openTextDocument(file);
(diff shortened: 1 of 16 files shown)


kanej:
Sorry, I merged the issue on linear into the original linear issue (#75), which triggered the close.

test/integration/helpers/misc.ts
```diff
@@ -3,6 +3,7 @@
 import * as path from "path";
```
kanej, line 3: What is your feeling on utility files like this?

On the one hand having a dumping ground is handy, on the other I am tempted to split out each key function into its own file named for the function.

How do you think we should deal with util files?

test/integration/helpers/files.ts
```diff
@@ -0,0 +1,70 @@
...
+  } else {
+    return deleteFile(file);
+  }
+};
+
+const createRandomFile = (
+  contents = "",
+  fileExtension = "txt"
+): Promise<vscode.Uri> => {
+  return new Promise((resolve, reject) => {
```
kanej, line 48: Might be worth using `fs/promises` here.

kanej (commented):
This is a great PR, great work!

I have added a few comments, but they are really about getting your thinking on style.

One thing to look into. When running the integration tests I get a warning about sym links/bad casing:

image

This may be a mac only issue (you can see it in the mac build logs). Any idea what is going there?

test/integration/helpers/misc.ts
antico5: I think I stopped the refactoring work halfway. I wanted to go from a single helper file to modules of helpers grouped by functionality. I can think of 3 approaches:
1 - move each function to a single file, without categorizing
2 - move each function a single file, categorizing them inside folders (i.e. `matchers/uriEqual.ts`, etc)
3 - have each category be a file with multiple exports (current approach for e.g. `helpers/files.ts`).

Which one do you like more ? I favor numbers 2 and 3 but the downside is they require thinking of some categorization/grouping and we may be tempted at some point to have an uncategorized/dump file.

kanej: The honest answer is, I prefer 2 but usually get to 2 via 3. I don't want to be committing to an abstraction too early (so dump in a broad category file - `helper.js` for the win), but if it clearly useful and widely used I start feeling guilty and want to move it out to its own file.

So like a coward, I am going to defer to your judgement on this one, either 2 or 3 in this case.

antico5:
@kanej Couldn't figure out the mac warning, but don't want to spend much more time on it since I have trouble reproducing it. 
I addressed the helpers refactor and the `fs` issue. I just replaced the async calls with the sync methods, they do the job and are simpler

kanej (approved):
Lets get this (squash) merged

PR #223, opened 2022-08-05
feat: Completions for msg.sender (#202)

### LINKED ISSUE #202: only default completions shown for `message.sender`
We have a report of `message.sender` not displaying the expected completions:

image

antico5:
I went with a simple approach of adding the address methods directly on the `msg.sender` global identifier. I think we don't have address type completions in place. 

Please advice on the approach I took for being able to identify the `msg.sender` node, since the existing code would only work for finding single word orphan nodes for globals (e.g. `msg`, `tx`, `abi`). But it wouldn't work for a chain of member access (e.g. `foo.bar.baz`. I did my best to avoid interfering with the existing logic for orphan nodes.

A possible improvement would be to move the logic from `getLastExpressionNode` and `getFullName` from the service module to the node classes themselves (`MemberAccessNode`, `IdentifierNode`, `Node`), but I'm not sure if I have the freedom to do that.

Closes #202

### DIFF
--- a/server/src/services/completion/onCompletion.ts
+++ b/server/src/services/completion/onCompletion.ts
@@ -13,6 +13,7 @@ import {
   Position,
   expressionNodeTypes,
   TextDocument,
+  MemberAccessNode,
 } from "@common/types";
 import { getParserPositionFromVSCodePosition } from "@common/utils";
 import { Logger } from "@utils/Logger";
@@ -127,7 +128,7 @@ export function doComplete(
 
     for (const orphanNode of documentAnalyzer.orphanNodes) {
       if (isNodePosition(orphanNode, newPosition)) {
-        definitionNode = orphanNode;
+        definitionNode = getLastExpressionNode(orphanNode);
         break;
       }
     }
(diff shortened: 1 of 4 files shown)


server/src/services/completion/onCompletion.ts
```diff
@@ -127,7 +128,7 @@ export function doComplete(
 
     for (const orphanNode of documentAnalyzer.orphanNodes) {
       if (isNodePosition(orphanNode, newPosition)) {
```
kanej, line 130: Nice catch here, I find the orphan nodes difficult to follow. It was only because your change forced me to read the comment that I got it:

```
  // Check if the definitionNode exists and if not, we will check if maybe Node exists in orphan Nodes.
  // This is important for "this", "super" and **global variables** because they exist only in orphanNodes.
 ```

Though I still don't have a good mental model of why global variables only exist in orphan nodes.

kanej (changes requested):
Well done, this is a minimum intervention in our usedef code - which is ideal. The only real comment is on `send`/`transfer`.

kanej (approved):
LGTM

PR #224, opened 2022-08-05
fix: Semicolon getting removed on virtual modifiers

### LINKED ISSUE #219: Formatter removes semicolon after the virtual modifier declaration
I have the following file:

```sol
// SPDX-License-Identifier: MIT
pragma solidity 0.8.2;

abstract contract Foo {
    modifier bar() virtual;
}

```

When I format it with `hardhat-vscode`, the formatter removes the semicolon after the modifier:
```diff
// SPDX-License-Identifier: MIT
pragma solidity 0.8.2;

abstract contract Foo {
-    modifier bar() virtual;
+    modifier bar() virtual
}

```

which results in invalid solidity code.

The issue seems to disappear when the modifier gets a default implementation:
```sol
// SPDX-License-Identifier: MIT
pragma solidity 0.8.2;

abstract contract Foo {
    modifier bar() virtual {
        _;
    }
}

```

antico5:
Solved by bumping prettier-plugin-solidity version to 1.0.0-dev.23

Closes #219


kanej (changes requested):
I forgot to mention that we use prettier in the server as well. Can you make sure the versions of prettier and the `prettier-solidity-plugin` match across client and server and they match Franco's new release `prettier-plugin-solidity@1.0.0-beta.24`.

kanej (approved):
LGTM!

PR #225, opened 2022-08-08
feat: Quickfix for specifying compiler version

### LINKED ISSUE #25: quick fix suggestion: add solidity pragma line
In response to __solidity 3420__ warning:

image

Provide a quick fix that uses the text of the warning to insert a pragma on the file top line.

antico5:
Closes #25 

We make use of the solc warning for code 3420 (Source file does not specify required compiler version! Consider adding "pragma solidity ^0.8.7;") to extract the pragma line and insert it into the contract code.

### DIFF
--- a/server/src/compilerDiagnostics/diagnostics/SpecifyCompilerVersion.ts
+++ b/server/src/compilerDiagnostics/diagnostics/SpecifyCompilerVersion.ts
@@ -1,0 +1,78 @@
+import { TextDocument } from "vscode-languageserver-textdocument";
+import {
+  CodeAction,
+  CodeActionKind,
+  Diagnostic,
+  Range,
+} from "vscode-languageserver/node";
+import { CompilerDiagnostic, ResolveActionsContext } from "../types";
+import { attemptConstrainToFunctionName } from "../conversions/attemptConstrainToFunctionName";
+import { HardhatCompilerError, ServerState } from "../../types";
+
+/**
+ * This diagnostic is shown when no compiler version is specified
+ * i.e. pragma solidity XXXX;
+ *
+ * The suggested quickfix adds the pragma solidity statement, by
+ * extracting the version from hardhat's warning message.
+ */
+export class SpecifyCompilerVersion implements CompilerDiagnostic {
+  public code = "3420";
+  public blocks: string[] = [];
+
+  public fromHardhatCompilerError(
+    document: TextDocument,
+    error: HardhatCompilerError
+  ): Diagnostic {
+    return attemptConstrainToFunctionName(document, error);
+  }
+
+  public resolveActions(
+    _serverState: ServerState,
+    _diagnostic: Diagnostic,
+    { uri, document }: ResolveActionsContext
+  ): CodeAction[] {
+    // Get the compiler specification code from hardhat warning
+    const regex = /pragma solidity .*;/;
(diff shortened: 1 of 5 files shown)


server/src/compilerDiagnostics/diagnostics/SpecifyCompilerVersion.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+export class SpecifyCompilerVersion implements CompilerDiagnostic {
+  public code = "3420";
+  public blocks: string[] = [];
+
+  public fromHardhatCompilerError(
+    document: TextDocument,
+    error: HardhatCompilerError
+  ): Diagnostic {
+    return attemptConstrainToFunctionName(document, error);
```
kanej, line 27: This will work but do an unnecessary check to see if the warning range covers a function (which it won't).

The idea of this function is that it is responsible for transforming the hardhat error (warning/error in reality) into a diagnostic (vscode concept), so it is a natural point to manipulate the range of the hardhat error; which we want to do when we have a whole function error/warning or a whole contract error/warning.

If you are not changing anything about the warning/error and are wanting the default transform you can put:

```javascript
import { passThroughConversion } from "@compilerDiagnostics/conversions/passThroughConversion";
...

public fromHardhatCompilerError(
    document: TextDocument,
    error: HardhatCompilerError
  ): Diagnostic {
    return passThroughConversion(document, error);
  }
```

server/src/compilerDiagnostics/diagnostics/SpecifyCompilerVersion.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    const checkRange: Range = {
+      start: _diagnostic.range.start,
+      end: {
+        line: _diagnostic.range.end.line,
+        character: _diagnostic.range.end.character + 1,
+      },
+    };
+
+    if (document.getText(checkRange) === "/") {
```
kanej, line 56: I think this covers the most common case, but I am tempted to say we tighten this up to do a check for whether it is a valid `SPDX-License-Identifier` line. If it is just a comment I think the rule should be that we put the compiler pragma on the first line and ignore the comment, only if the first line is a `SPDX-License-Identifier` line does the compiler pragma move from first to second line.

server/test/services/codeactions/specifyCompilerVersion.ts
```diff
@@ -0,0 +1,65 @@
...
+  before(async () => {
+    testContractText = (
+      await fs.promises.readFile(
+        path.join(__dirname, "testData", "SpecifyCompilerVersion.sol")
+      )
+    ).toString();
+  });
+
+  describe("Specify compiler version", () => {
+    it("adds pragma solidity line", async () => {
```
kanej, line 19: Nice, could we get two further tests, so we end up with:

* no spdx license line
* a spdx license line
* no spdx license line, but a standard comment on the first line

That way if someone is trying to figure out whether the behaviour around add pragma lines when the first line is a comment vs a license line, we have an explicit test to demonstrate what we intended the rule to be.

kanej (changes requested):
Awesome!

I have added comments for the stuff that came out of the vscode call, basically upping the comment check.

Having both `add pragma` and `add license identifier` is going to be superb.

kanej (approved):
Brilliant. Lets get this merged!

PR #228, opened 2022-08-10
docs: update publish instructions

kanej:
The publish instructions need updated to follow acutal usage and include deploying to open vsix.


kanej:
I updated these instructions based what was needed to get the last release out. I propose we merge and use them for the next, and separate out any automation scripts into their own PR. In particular I would like to see a deploy script that packages and pushes to openvsx and the microsoft marketplace.

antico5 (approved):
Looks good to me !

PR #230, opened 2022-08-10
feat: Quickfix for partially-specified multi override

### LINKED ISSUE #50: quickfix suggestion: missing contract specifier on override
If you specify an override but miss a contract specifier you get `solidity 4327` as the error. We should add a quickfix to resolve.

## Example

image

antico5:
Closes #50


kanej (changes requested):
This is excellent. I was worried that this was going to get tricky but you have put together a clean implementation.

In running it through I was pondering whether we should change location of the error from `override(A,B)` if there is a missing contract to the function name. My instinct was to be consistent with where the other quickfixes are; but I think you made the right call on that by leaving the warning on the override keyword. It has better locality to the issue.

If we can get the contracts sorted (scope creep I know), this looks good to go.

kanej (approved):
LGTM!

PR #232, opened 2022-08-11
fix: Open files not being analyzed because of validator starting

### LINKED ISSUE #229: inline validation not appearing until edit
Reported on discord: https://discord.com/channels/750408878008827925/951537942239797338/1006713748603539567

Inline validation on the open editor file, doesn't start appearing until an edit happens on the file.

On starting the extension, indexing kicks in but the open editor window goes to a `solidity language status` error, indicating that validation is still starting:

image

I was able to reproduce using this repo: https://github.com/thiras/altus

antico5:
Closes #229

Open files were being sent for validation before the validation workers were operational.

I introduced a new notification sent from the server for each validation worker that got up and running. The client will revalidate files within the project associated to each given worker.

### DIFF
--- a/server/src/services/validation/HardhatWorker.ts
+++ b/server/src/services/validation/HardhatWorker.ts
@@ -2,6 +2,7 @@ import * as childProcess from "child_process";
 import * as path from "path";
 import { HardhatProject } from "@analyzer/HardhatProject";
 import { Logger } from "@utils/Logger";
+import { Connection } from "vscode-languageserver";
 import {
   InitialisationCompleteMessage,
   InvalidatePreprocessingCacheMessage,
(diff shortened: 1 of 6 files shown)


server/src/services/validation/HardhatWorker.ts
```diff
@@ -93,6 +97,9 @@ export class HardhatWorker implements WorkerProcess {
             this.logger.trace(
               `initialisation complete for ${this.project.basePath}`
             );
+            this.connection.sendNotification("custom/worker-initialized", {
+              project: this.project,
```
kanej, line 101: Let's change the passed object to `{projectBasePath: string}`, in case we add something non-serializable to project without realizing it and break this.

kanej (changes requested):

kanej (approved):
I did a quick test on windows, and the fix looks good. LGTM!

PR #237, opened 2022-08-18
feat(analyzer): add support for remappings.txt

fubhy:
Hi. Sorry for not opening an issue beforehand. I discussed this with @fvictorio on Discord a bit earlier today and thought I'd go ahead and experiment a bit.

This adds support for `remappings.txt` on a per-project basis. It looks for remappings.txt files beides hardhat config files. 

This is probably not the ideal solution but there's no clear documentation from dapptools or foundry where these remappings files are supposed to be and how they are supposed to be interpreted (e.g. are they supposed to be next to a project config file like hardhat.config, foundry.toml, etc.)? 

Generally, it seems like foundry is actually moving away from remappings.txt towards having those defined in the foundry.toml directly. In that case, the best solution might be to look for foundry.toml files and run the `forge remappings` commands to have them resolved by forge instead of attempting to do that ourselves.

### DIFF
--- a/server/src/parser/analyzer/HardhatProject.ts
+++ b/server/src/parser/analyzer/HardhatProject.ts
@@ -1,4 +1,4 @@
-import { ISolProject } from "@common/types";
+import { ISolProject, Remapping } from "@common/types";
 import { WorkspaceFolder } from "vscode-languageserver-protocol";
 
 export function isHardhatProject(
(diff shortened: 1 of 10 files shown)


aathan:
Please note that this code was then refactored into separated support in `FoundryProject` and that mixed Foundry/Hardhat projects, per https://hardhat.org/hardhat-runner/docs/advanced/hardhat-and-foundry, are now supported via:

```
npm install --save-dev @nomicfoundation/hardhat-foundry
```

and

```
import "@nomicfoundation/hardhat-foundry";
```

without use a `preprocessor:` in your hardhat config.

PR #239, opened 2022-08-20
feat: add `forge fmt` option

fubhy:
Adapting https://github.com/juanfranblanco/vscode-solidity/pull/359 to work with this extension too (h/t @shekhirin)

### DIFF
--- a/client/src/formatter/forgeFormatter.ts
+++ b/client/src/formatter/forgeFormatter.ts
@@ -11,19 +11,18 @@ export async function formatDocument(
     lastLine.range.end
   );
 
-  const p = cp.execFile("forge", [
-    "fmt",
-    "--check",
-    "--raw",
-    document.uri.fsPath,
-  ]);
+  const formatted = await new Promise<string>((resolve, reject) => {
+    const forge = cp.execFile("forge", ["fmt", "--raw", "-"], (err, stdout) => {
+      if (err !== null) {
+        return reject(err);
+      }
 
-  let formatted = "";
-  if (p.stdout) {
-    for await (const chunk of p.stdout) {
-      formatted += chunk;
-    }
-  }
+      resolve(stdout);
+    });
+
+    forge.stdin?.write(document.getText());
+    forge.stdin?.end();
+  });
 
   return [vscode.TextEdit.replace(fullTextRange, formatted)];
 }
--- a/client/src/formatter/forgeFormatter.ts
+++ b/client/src/formatter/forgeFormatter.ts
@@ -1,0 +1,29 @@
+import * as vscode from "vscode";
+import * as cp from "child_process";
+
+export async function formatDocument(
+  document: vscode.TextDocument
+): Promise<vscode.TextEdit[]> {
+  const firstLine = document.lineAt(0);
+  const lastLine = document.lineAt(document.lineCount - 1);
+  const fullTextRange = new vscode.Range(
+    firstLine.range.start,
+    lastLine.range.end
+  );
+
+  const p = cp.execFile("forge", [
+    "fmt",
+    "--check",
+    "--raw",
+    document.uri.fsPath,
+  ]);
+
+  let formatted = "";
+  if (p.stdout) {
+    for await (const chunk of p.stdout) {
+      formatted += chunk;
+    }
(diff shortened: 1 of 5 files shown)


client/src/formatter/forgeFormatter.ts
```diff
@@ -0,0 +1,29 @@
...
+  document: vscode.TextDocument
+): Promise<vscode.TextEdit[]> {
+  const firstLine = document.lineAt(0);
+  const lastLine = document.lineAt(document.lineCount - 1);
+  const fullTextRange = new vscode.Range(
+    firstLine.range.start,
+    lastLine.range.end
+  );
+
+  const p = cp.execFile("forge", [
```
kanej, line 14: This will format the file on disk, but format is called on the modified in-memory buffer of the open file.

Looking at the `forge fmt` options, passing `-` as the file makes it read from stdin, so maybe something like:

```ts
const forgePath = path.resolve(os.homedir(), ".foundry/bin/forge");

    const forge = cp.execFile(
      forgePath,
      ["fmt", "--raw", "-"],
      (err, stdout) => {
        if (err !== null) {
          return reject(err);
        }

        resolve(stdout);
      }
    );
    const text = document.getText();
    forge.stdin?.write(text);
    forge.stdin?.end();
  });

  return [vscode.TextEdit.replace(fullTextRange, formatted)];
```

My attempts kept on failing until I actually read that `--check` was triggering a fail on bad formatting, so in this usage we should leave it out.

kanej (changes requested):
This is great.

Reading from in-memory is the only block.

We are currently working on better integration tests, and this will be a good excuse to improve our formatting tests. But we will handle that in a later PR.

I am really excited to get some foundry support in. Thanks!

client/src/formatter/forgeFormatter.ts
fubhy: Thanks for the feedback @kanej. Using `forge fmt` with the `--raw` and `--check` flag indeed returns a non-zero exit code in case formatting is required but it does not write to the file in that case (because of the `--check` flag).

So in combination, `--raw` and `--check` will cause the full formatted file to be printed to stdout instead of modifying the file in place. So imho this is the desired behavior (as it is implemented now). No?

kanej: Ah sorry, I have added confusion with the comment above, let me try and clarify.

To get formatting working on the in-memory open document, I suggested the code snippet above, which pipes the current documents text through `forge fmt` via stdin.

It changes the usage of execFile using a callback and references to stdin. For that code snippet, if the `--check` flag is included, then any text passed in that is not perfectly formatted will cause the exit code of `forge fmt` to be non-zero, and execFile returns an error in the callback. The snippet above would then rethrow that error via the promise, rather than return the formatted text edit. It is the code snippets changed usage of execFile and its error handling that forces the change to remove `--check`.

Put another way the snippet above would error on most inputs instead of returning a `vscode.TextEdit.replace` result if `--check` is included. But when we are piping from stdin to stdout without touching the filesystem, we don't need to return a non-zero exit code, so we can leave `--check` out.

fubhy: Ah, I see... So lapidary, in order to get it to work for `> Format` (without saving) or `> Format on keystroke`, we need to be able to format "in-memory". I misinterpreted that comment indeed. I can adjust that then unless @shekhirin beats me to it. Gotta clean up some toddler vomit first. Not figuratively speaking.

kanej: Toddler vomit should always take priority, no rush!

fubhy: @kanej Adjusted it according to your recommendation (minus the `forge` bin change, I left some comments in Discord for that)

kanej: Hey @fubhy, I am not seeing any changes in the PR? Or am I missing something?

fubhy: @kanej Sorry, pushed to the wrong remote ;-)

kanej (approved):
I have been able to test the new `forge fmt` option on windows.

PR #241, opened 2022-08-23
chore: link to error file on status item (#238)

### LINKED ISSUE #238: Validation is blocked on the entire project when there's an invalid import on a single file.
As the title says, if at least 1 file has an invalid import (either relative or library import), the validation gets blocked for all files in the project. The expected behavior is that the diagnostic is shown for the file with the error, and other files are validated normally.

antico5:
Added a link on the status item linking to the file that is importing a non existent file.

The way hardhat works, is that the project cannot be compiled if the dependency graph cannot be built successfully (e.g. on an incorrect import). The extension already shows a diagnostic if the import line is on the current file, but just a status item if the error is on another file. This change adds a link to that status item

Closes #238

### DIFF
--- a/client/src/popups/setupValidationJobHooks.ts
+++ b/client/src/popups/setupValidationJobHooks.ts
@@ -1,3 +1,4 @@
+import path from "path";
 import { LanguageStatusItem, languages, LanguageStatusSeverity } from "vscode";
 import { LanguageClient } from "vscode-languageclient/node";
 import { ExtensionState } from "../types";
@@ -13,6 +14,7 @@ export interface ValidationJobFailureNotification {
   reason: string;
   displayText: string;
   projectBasePath: string;
+  errorFile?: string;
 }
 
 export type ValidationJobStatusNotification =
@@ -65,6 +67,18 @@ function updateValidationStatusItem(
     ? "Solc Version"
     : notification.displayText;
   statusItem.busy = false;
+
+  if (!notification.validationRun && notification.errorFile !== undefined) {
+    statusItem.command = {
+      title: "Open file",
+      command: "vscode.open",
+      arguments: [
+        path.join(notification.projectBasePath, notification.errorFile),
+      ],
+    };
+  } else {
+    statusItem.command = undefined;
+  }
 
   return statusItem;
 }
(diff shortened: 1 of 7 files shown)


kanej:
I did a run through with the file link. It is a clear improvement, but the error on the import line won't appear until an edit has happened on the page.

I think we need to resolve it so that jumping to the dependency file with the error, displays the error.

We could:
1. trigger validation on open editor change (not sure which vscode event this is)
2. change diagnostics so that we display errors/warnings on dependent files but also clean them up when appropriate (there is a subtelty here between the source paths we get back from solc on a successful validation, and the file paths that could lead to a hardhat import line error - these can be outside the dependency graph of the current file)

1 is more straightforward to implement. I think 2 we are going to have to do at some point.

client/src/popups/setupValidationJobHooks.ts
```diff
@@ -66,5 +68,17 @@ function updateValidationStatusItem(
     : notification.displayText;
   statusItem.busy = false;
 
+  if (!notification.validationRun && notification.errorFile !== undefined) {
+    statusItem.command = {
+      title: "Open file",
+      command: "vscode.open",
+      arguments: [
+        path.join(notification.projectBasePath, notification.errorFile),
```
kanej, line 76: The navigation fails on windows and I suspect it is this line.

kanej: I suspect we need to pass the result of `path.join` to `decodeUriAndRemoveFilePrefix()`

server/src/services/validation/validate.ts
```diff
@@ -132,11 +135,18 @@ function hardhatThrownFail(
 
     const displayText = hardhatError.errorDescriptor.title;
 
+    const errorFile = isHardhatSourceImportError(hardhatError)
+      ? decodeUriAndRemoveFilePrefix(
```
kanej, line 139: Ugh, I tested on windows and we need to decode Uri and add a prefix of `/` on windows (i.e. `/c:/projects/example.sol` rather than `c:/project/example.sol`).

So maybe something like:
```typescript
function resolveErrorFilePath(projectBasePath: string, hardhatError: any): string {
  const errorPath = decodeUriAndRemoveFilePrefix(
    path.join(projectBasePath, hardhatError.messageArguments.from)
  )

  const osPath = os.platform() === "win32" ? `/${errorPath}` : errorPath

  return osPath
}
```

kanej (approved):

kanej:
I was able to test the link on windows. LGTM!

PR #253, opened 2022-09-08
fix: diagnostics wrong position on unicode characters (#248)

### LINKED ISSUE #248: Incorrect range on diagnostic on contracts with non-ascii characters
Solc references offsets as byte count instead of unicode character count, but VSCode uses the latter. This results in the diagnostics being shown in incorrect ranges.

antico5:
Closes #248


server/src/services/validation/worker/build/solcCompile.ts
```diff
@@ -43,5 +43,28 @@ export async function solcCompile(
...
 
+  // Normalize errors' sourceLocation to use utf-8 offsets instead of byte offsets
+  for (const error of output.errors) {
+    const source = input.sources[error.sourceLocation?.file];
+
+    if (source === undefined) {
+      continue;
+    }
+
+    const normalizeOffset = (text: string, offset: number) => {
```
alcuadrado, line 54: I'd be careful about the performance of this.

Also, while the solc output is in `utf-8` (I checked with them), the file may not. You can change the file encoding in the vscode UI to something different and see what happens. For example, try utf-16 and include some surrogate-pair chars.

antico5: Just tested the performance. On a big contract (https://github.com/NexusMutual/smart-contracts/blob/nexus-v2/contracts/modules/capital/Pool.sol), the decoding + encoding of the whole file takes 0.068ms, allowing ~15k operations per second. Personally I wouldn't add complexity to make it faster but if needed it could be optimized a bit for example in cases where a single file has multiple errors. 

I'll test other encodings on vs code and see how we can best support them

antico5: I tried with multiple encodings (i.e. utf16), and all works well. Looks like VSCode is performing the necessary decoding/encoding under the hood, and when the contract code reaches solc it's already correctly encoded. I added another integration test for this.

server/src/services/validation/worker/build/solcCompile.ts
```diff
@@ -43,5 +43,32 @@ export async function solcCompile(
...
```
kanej, line 54: Can we move this function out and declare it in the top level file?

kanej (approved):
LGTM!

PR #257, opened 2022-09-09
feat: auto insert semicolon on import completion

### LINKED ISSUE #104: Auto insert ';' at least on import
When completing an import: `import "./X"`, we should auto append the `;` to the end of the line. Our parsing is disrupted by the lack of a `;`.

antico5:
Closes #104 

When accepting a completion from imports, a semicolon will be appended if the current line doesn't end with a semicolon already.

### DIFF
--- a/server/src/services/initialization/onInitialize.ts
+++ b/server/src/services/initialization/onInitialize.ts
@@ -38,7 +38,7 @@ export const onInitialize = (serverState: ServerState) => {
         textDocumentSync: TextDocumentSyncKind.Incremental,
         // Tell the client that this server supports code completion.
         completionProvider: {
-          triggerCharacters: [".", "/", '"'],
+          triggerCharacters: [".", "/", '"', "'"],
         },
         signatureHelpProvider: {
           triggerCharacters: ["(", ","],
--- a/test/integration/index.ts
+++ b/test/integration/index.ts
@@ -3,12 +3,20 @@ import path from "path";
 import Mocha from "mocha";
 // eslint-disable-next-line import/no-extraneous-dependencies
 import glob from "glob";
+import { sleep } from "./helpers/sleep";
 
 export function run(): Promise<void> {
   // Create the mocha test
   const mocha = new Mocha({
     ui: "tdd",
     color: true,
+    rootHooks: {
+      beforeAll: async () => {
+        await sleep(5000); // Wait for the extension to be loaded
+      },
+    },
+    timeout: 30000,
+    retries: 1,
   });
 
   const testsRoot = path.resolve(__dirname, "tests");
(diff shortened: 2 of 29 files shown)


server/src/services/initialization/onInitialize.ts
```diff
@@ -38,7 +38,7 @@ export const onInitialize = (serverState: ServerState) => {
         textDocumentSync: TextDocumentSyncKind.Incremental,
         // Tell the client that this server supports code completion.
         completionProvider: {
-          triggerCharacters: [".", "/", '"'],
+          triggerCharacters: [".", "/", '"', "'"],
```
kanej, line 41: Can you walk me through the cases this supports?

test/integration/index.ts
```diff
@@ -3,12 +3,20 @@ import path from "path";
...
+import { sleep } from "./helpers/sleep";
 
 export function run(): Promise<void> {
   // Create the mocha test
   const mocha = new Mocha({
     ui: "tdd",
     color: true,
+    rootHooks: {
+      beforeAll: async () => {
+        await sleep(5000); // Wait for the extension to be loaded
```
kanej, line 15: What is your feeling around sleeps in the integration tests suite?

Is this just a necessary evil as we don't have other options? How brittle do you think we are on our integration tests?

server/src/services/initialization/onInitialize.ts
kanej: Is this to support `import './Auth.sol';` for import statements?

kanej (commented):
This is looking good and I think will make users smile if they notice.

I have left a few comments, but the trigger characters inclusion of `'` is the only thing I want to double check.

server/src/services/initialization/onInitialize.ts
antico5: So I noticed that I wasn't getting import completion, and it was because I use single quotes instead of double, because it's a single keystroke instead of a chord. Even though the formatter may change it to double quotes afterwards, I'll always use single quotes, and thought that this may be happening to a lot of people.

test/integration/index.ts
antico5: This `sleep` call only runs once before all the test suites, ensuring that they will have the features available. If there was a way to check that the LSP is fully initialized I'd use that but I didn't find a way to read extension state from the vscode API. Maybe I just didn't search enough but this does the trick.

server/src/services/initialization/onInitialize.ts
kanej: Based on the solidity grammar, it looks as if import paths are just `non-empty-string-literals` and so can be either single or double quoted, and that has just been missed previously.

If we are going to fix that as part of this PR we should probably update the completion guard here: https://github.com/NomicFoundation/hardhat-vscode/blob/f8d9498a4f386484218647eb18b86c0aa99e7158/server/src/services/completion/onCompletion.ts#L138

test/integration/index.ts
kanej: If the integration tests are reliable enough with this setup, lets hold off on finding other solutions.

kanej (approved):
LGTM!

PR #259, opened 2022-09-14
feat: analysis for pure foundry projects

antico5:
Identify folders containing `foundry.toml` files as foundry projects. Ignores directories under `lib` by default.


server/src/services/initialization/indexWorkspaceFolders.ts
```diff
@@ -141,65 +179,44 @@ function parseRemappings(rawRemappings: string, basePath: string) {
...
+  workspaceFileRetriever: WorkspaceFileRetriever
+) {
   const uri = decodeUriAndRemoveFilePrefix(workspaceFolder.uri);
-  const hardhatConfigFiles = await workspaceFileRetriever.findFiles(
-    uri,
-    "**/hardhat.config.{ts,js}",
-    ["**/node_modules/**"]
-  );
+  return workspaceFileRetriever.findFiles(uri, "**/foundry.toml", [
+    "**/lib/**",
```
kanej, line 188: Ah good catch

test/integration/tests/remappings/remappings.test.ts
```diff
@@ -9,63 +9,73 @@ import {
...
-
-    await vscode.commands.executeCommand("editor.action.goToDeclaration");
-
-    assert.equal(getCurrentEditor().document.fileName, otherImportedUri.fsPath);
-    assertPositionEqual(
-      getCurrentEditor().selection.active,
-      new vscode.Position(4, 9)
-    );
-  });
+  for (const project of ["hybrid", "pure_foundry"]) {
```
kanej, line 12: Does this affect debugging at all?

server/src/services/initialization/indexWorkspaceFolders.ts
```diff
@@ -23,79 +29,111 @@ export interface IndexWorkspaceFoldersContext {
...
+        buildProject<FoundryProject>(
+          FoundryProject,
+          configFile,
+          workspaceFolder,
+          workspaceFileRetriever
+        )
+      )
+    );
+
+    for (const project of [...hardhatProjects, ...foundryProjects]) {
```
kanej, line 102: I am trying to remember, this is a guard against projects nested in projects?

Does this need to consider foundry in hh or hh within foundry? Or are we sticking to a rule that says, only the top level project counts?

test/integration/tests/remappings/remappings.test.ts
antico5: If you wanted to use breakpoints or other debugging tools you may need to use conditions for the specific project. But I think it's a good tradeoff since otherwise too much code would be duplicated

server/src/services/initialization/indexWorkspaceFolders.ts
kanej: Lets deal with that when we do the refactor.

kanej (approved):

antico5:
This PR is deprecated after the framework-agnostic refactor

PR #260, opened 2022-09-16
fix: trying to validate non-hardhat contracts

### LINKED ISSUE #255: validation error when hardhat task not passed file parameter
In sentry we see validation errors due to a missing parameter being passing into one of the hardhat tasks:

```json
{
  "errorDescriptor": {
    "description": "You tried to run a task, but one of its required arguments was missing. \n\nPlease double check how you invoked Hardhat or ran your task.",
    "message": "The '%param%' parameter expects a value, but none was passed.",
    "number": 306,
    "shouldBeReported": false,
    "title": "Missing task argument"
  },
  "messageArguments": {
    "param": "file"
  },
  "name": "HardhatError"
}
```

There is only one hardhat task that takes a file parameter: compile (https://github.com/NomicFoundation/hardhat/blob/master/packages/hardhat-core/src/builtin-tasks/compile.ts#L218)

## TODO

- [ ] Figure out how to reproduce this error
- [ ] Fix it

antico5:
The reported error on (#255) happens when you are on a hardhat project, but open a contract that's not inside the configured contracts folder (`contracts` by default). This PR limits what files can go through the validation pipeline. The contracts folder is identified on worker initialization, since it loads the hardhat config. It's passed back to the server process on the initialization complete message. There's no race condition since validation is blocked anyways before worker is fully initialized.

Closes #255


kanej:
Based on discussion with @alcuadrado, we want to edit the validation within the worker to add a pre-compile task check to determine whether the current file path is in the dependency graph (that it is a hardhat project file). If it is not we return an error and display a useful error message. @armando

antico5:
> Based on discussion with @alcuadrado, we want to edit the validation within the worker to add a pre-compile task check to determine whether the current file path is in the dependency graph (that it is a hardhat project file). If it is not we return an error and display a useful error message. @Armando

What error message should we show and how should we display it ? (I'm thinking status bar item but want to be sure)

kanej:
> > Based on discussion with @alcuadrado, we want to edit the validation within the worker to add a pre-compile task check to determine whether the current file path is in the dependency graph (that it is a hardhat project file). If it is not we return an error and display a useful error message. @Armando
> 
> What error message should we show and how should we display it ? (I'm thinking status bar item but want to be sure)

Status bar item for the message - like our other "problem with validation process" type errors. There is no import line to link it back to so no sense in which we can have a diagnostic.

For the message, maybe something like "Not part of Hardhat project". Do you think adding a link to the docs explaining the layout would work?

antico5:
@kanej Updated PR

server/src/services/validation/worker/build/buildInputsToSolc.ts
```diff
@@ -101,6 +102,10 @@ export async function buildInputsToSolc(
 
   getValidationFile(workerState, buildJob);
 
+  if (buildJob.context.file === undefined) {
```
kanej, line 105: Is it 1 to 1 that all build jobs without a context file are outside the hh proj?

antico5: This condition happens when the file is neither on the source paths or a dependency of one of them

PR #261, opened 2022-09-16
fix: open hardhat.config fails on vscode web (#227)

### LINKED ISSUE #227: Open hardat.config.{js,ts} fails in vscode web
If you click on the solidity status bar item and then click the hardhat config link to navigate to the config file, you get an error if you are in vscode for web.

image

antico5:
The command `vscode.open` has better multi-platform support when passing a URI object instead of a path string.

Tested on linux, windows and gitpod.

Closes #227


kanej (approved):
LGTM!

PR #267, opened 2022-09-30
feat: projectless validation

antico5:
The scope of this PR is to provide validation to hardhat files not belonging to a hardhat project. This includes files in directories not belonging to hardhat, as well as random solidity files open in the editor.

Besides handling these cases, the PR involves the implementation of a particular abstraction between the LSP and "framework providers", to facilitate the inclusion of other providers in the future, while reusing the most code on the LSP side. The "Projectless" implementation works as an adapter as well.

Some concerns that may come up and why I implemented them that way:
- On `ProjectlessCompilationBuilder._crawlDependencies`, there are calls to the analyzer. This can be refactored or a better abstraction could be in place. In theory, indexing and analysis happens before validation, but this was needed because you can open a random solidity file that has dependencies, but be on a completely different folder in your workspace.
- Lookup of hardhat.config file bundled with the extension. Depending whether you are running the extension in dev mode, running unit tests, or running from a final packaged version, the mandatory hardhat.config file that we need to invoke hardhat tasks can be located in different places, so there's some lookup logic for that in `CompilationService`
- ProjectlessIndexer is not called from `indexWorkspaceFolders`, as we planned to. When discussing this, ideally we wanted the LSP to not know about any adapter, even the Projectless one, but on a scenario where you open an external solidity file (not in workspace), it was not analyzed and we have no project associated to it. So I solved this by having ProjectlessProject be the kind of fallback project when no other project is associated to a sol file.
- `OutputConverter` is duplicating some of the logic present in other parts of the code. Eventually this should be refactored into having a single source of logic for validation results, including both framework-specific and agnostic results.
- A new flag was introduced to prevent validation happening before indexing finished. This is because when a file is opened, it's attempted to be validated. If we did this before indexing finishes, a file can be identified as Projectless and be validated using that strategy. But after indexing, it could turn out that it was a hardhat contract so it was meant to be validated differently, and for that amount of time it could show wrong diagnostics (i.e. unexisting import).


scripts/bundle.js
```diff
@@ -54,13 +54,14 @@ async function main() {
...
       "./server/out/index": "./server/src/index.ts",
+      "./server/out/hardhat.config": "./server/src/hardhat.config.ts",
       "./server/out/worker": "./server/src/services/validation/worker.ts",
     },
     bundle: true,
     minifyWhitespace: true,
     minifyIdentifiers: false,
     minifySyntax: true,
-    external: ["vscode", "@nomicfoundation/solidity-analyzer"],
+    external: ["vscode", "@nomicfoundation/solidity-analyzer", "fsevents"],
```
kanej, line 64: I can't find any other references to fsevents

server/package.json
```diff
@@ -61,10 +63,12 @@
     "got": "^11.8.2",
     "hardhat": "^2.6.0",
     "js-yaml": "^4.1.0",
+    "lodash": "^4.17.21",
```
kanej, line 66: I am not opposed to lodash, but it does change what is idiomatic in the codebase. Put another way you can get really function in style (which I like). Lets confirm we are happy to take this turn with @alcuadrado

server/src/services/initialization/onInitialize.ts
```diff
@@ -30,6 +31,11 @@ export const onInitialize = (serverState: ServerState) => {
       extensionVersion,
     });
 
+    // fetch available solidity versions
+    await getAvailableSolcVersions(serverState);
```
kanej, line 35: I would change the name to `setAvailableSolcVersions` or alter it to `serverState.solcVersions = await getAvailableSolcVersions();`

server/src/services/initialization/onInitialized.ts
```diff
@@ -33,23 +34,24 @@ export const onInitialized = (
...
     }
 
+    // index folders
     await serverState.telemetry.trackTiming("indexing", async () => {
       await indexWorkspaceFolders(
         { ...serverState, workspaceFolders: [] },
         workspaceFileRetriever,
         serverState.workspaceFolders
       );
+      serverState.indexingFinished = true;
```
kanej, line 44: What do you think about moving indexing to initialize. Solving this the same way we solve the `validator still staring` error?

kanej: Agreed.

scripts/bundle.js
kanej: Ah, that makes sense.

antico5:
This PR is deprecated after the framework-agnostic refactor. Projectless was included and updated with a few changes on the big refactor

PR #275, opened 2022-11-02
Update README.md

joejordan:
Clarify that Hardhat needs to be setup on the project for this plugin to work.


kanej:
Hey @joejordan, we have just released in `v0.6.0` experimental support for foundry projects, both as standalone projects and as hybrid hardhat/foundry setup.

If that is something you want to try in your project, any feedback/bug reports would be greatly appreciated.

kanej:
I am going to close this as we now support solidity files outside of a hh project and are working through our foundry support.

PR #276, opened 2022-11-04
async analysis

antico5:
This changes our projects' `resolveImportPath()` , and everything related to import resolution involved in the parser, to be async. This allows us to use hardhat on import resolution. What this enables us in the short term is supporting the plugin `hardhat-foundry`, which makes use of that async import transforming task.

### DIFF
--- a/server/src/frameworks/Hardhat/worker/WorkerProtocol.ts
+++ b/server/src/frameworks/Hardhat/worker/WorkerProtocol.ts
@@ -9,6 +9,8 @@ export enum MessageType {
   ERROR_RESPONSE,
   FILE_BELONGS_REQUEST,
   FILE_BELONGS_RESPONSE,
+  RESOLVE_IMPORT_REQUEST,
+  RESOLVE_IMPORT_RESPONSE,
   BUILD_COMPILATION_REQUEST,
   BUILD_COMPILATION_RESPONSE,
   INITIALIZATION_FAILURE,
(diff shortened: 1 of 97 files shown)


server/src/frameworks/Hardhat/worker/WorkerProtocol.ts
```diff
@@ -66,6 +68,26 @@ export class FileBelongsResponse extends ResponseMessage {
   }
 }
 
+export class ResolveImportRequest extends RequestMessage {
```
kanej, line 71: In typescript I think it is more idiomatic to use a type if it is just a data carrier you want.

server/src/parser/analyzer/nodes/ImportDirectiveNode.ts
```diff
@@ -22,28 +21,24 @@ export class ImportDirectiveNode extends AbstractImportDirectiveNode {
...
-        solFileEntry.project.resolveImportPath(
-          this.realUri,
-          importDirective.path
-        ) ?? "";
-    } catch (err) {
-      this.uri = "";
-    }
+    this.uri = resolvedUri;
+
+    // try {
```
kanej, line 32: Lets remove the comment.

kanej (approved):

server/src/frameworks/Hardhat/worker/WorkerProtocol.ts
antico5: After giving it some thought, you are right that in this case classes are being used incorrectly. These objects are being serialized/deserialized through the ipc between parent and child processes. So they are acting as just plain objects, losing their prototype. I'll change this on the other big PR and on this one

PR #278, opened 2022-11-10
feat: framework agnostic import completions + foundry implementation

antico5:
The `onCompletion` service has been refactored to be framework-independant (relative imports are still there). Hardhat imports through node_modules have been moved behind the hardhat adapter and a new one was created for foundry. It works by providing extra import completions based on remappings. Integration tests have been added for this.


server/test/services/completion/imports.ts
```diff
@@ -98,42 +98,6 @@ describe("Parser", () => {
             errors: [],
           }));
         });
-
-        it("should list sol files under a package folder in node_modules", () =>
```
kanej, line 102: Is the idea here to swap the unit tests for the integration tests?

server/src/frameworks/Foundry/ImportCompleter.ts
```diff
@@ -0,0 +1,71 @@
+import {
+  CompletionItem,
+  CompletionItemKind,
+  Position,
+} from "vscode-languageserver-types";
+import { SolFileIndexMap } from "../../parser/common/types";
+import { Remapping } from "./Remapping";
+
+export class ImportCompleter {
```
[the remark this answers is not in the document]
antico5: True, it can be done both ways :sweat_smile: but this is the one that comes naturally to me, and I thought that it'd be appropiate since this part of the code uses classes as well due to the recent refactor. But I'm open to change the style of programming if you think it's more appropiate/desirable for this project. In any case, I'd like to stick to just one approach. Either using mostly classes, or all plain objects with functions that operate on them.

server/test/services/completion/imports.ts
antico5: The removed unit tests were related to hardhat import completions, so I thought that instead of mocking the necessary components, I'd just write integration tests which ensure that ensure the whole completion process works as expected. They slower so that's definitely a tradeoff, but I'm not running them all the time. Should I restore the unit tests and write the required mocking components to support it ?

kanej: I would have an integration for the happy path and unit tests for the edge cases (the feature works across platform and here is my definition of works in detail).

If we only leverage the integration test suite it will get slower and slower over time and we miss out on the design pressure that comes at the component level from having to write components that can be tested.

server/src/frameworks/Foundry/ImportCompleter.ts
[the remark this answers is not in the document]
kanej: I think that is a false dichotomy; we can leverage classes and have have standalone functions in one code base.

What we need is a convention on when we would use a class and when we would use a standalone function.

The framework adapter is a good example of where a class is justified (multiple implementations to same interface, need to combine state + logic). It pays for itself as an abstraction.

The `ImportCompleter` doesn't need inheritance, different implementations aren't going to be used from within the Foundry adapter, the state doesn't have a different lifetime from individual function calls; it is just a holder for the `ImportComplete` process.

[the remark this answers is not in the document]
antico5: Makes sense. I'll try to better distinguish when to use which :+1:

server/test/services/completion/imports.ts
kanej: We got our answer on testing (add a new integration test suite against the LSP). So the question here is, are there any test cases in these unit tests that we think should make the migration to LSP integration tests - and hence should be ported to ui integration tests? Or do we think we have coverage of the import paths, in which case we are good?

antico5: I'd say we are good, for this case I wrote at least 1 e2e test for each scenario removed from unit tests. After we have LSP integration tests I'll work on adding more thorough coverage since we'll have a much faster test suite.

kanej: Brilliant. Then let me approve this.

kanej (approved):
LGTM!

PR #280, opened 2022-11-17
optimization: disable optimizer and outputSelection

antico5:
This change makes the compilation 3x-10x faster, specially for hardhat projects.

TBD if this has any negative implication

### DIFF
--- a/server/src/services/validation/CompilationService.ts
+++ b/server/src/services/validation/CompilationService.ts
@@ -12,6 +12,9 @@ export class CompilationService {
     const hre = this._getHRE();
     const { input } = compilationDetails;
 
+    // This makes the compilation up to 10x faster
+    delete (input.settings as any).outputSelection;
+
     // Find or download solc compiler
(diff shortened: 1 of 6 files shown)


server/src/services/validation/CompilationService.ts
```diff
@@ -12,6 +12,10 @@ export class CompilationService {
     const hre = this._getHRE();
     const { input } = compilationDetails;
 
+    // This makes the compilation up to 10x faster
+    delete (input.settings as any).optimizer;
```
kanej, line 16: What was the final decision on this one?

I thought we were held to keeping it due to changes to semantics?

antico5: I'm not sure, do we have a counter example for that ? In my opinion this optimization is a huge QoL booster. The delay between typing and getting diagnostics is very noticeable.

kanej (approved):
LGTM!

PR #281, opened 2022-11-17
feat: quickfix for data location

antico5:
This feature adds a quickfix for solidity errors involving data location (add/change/remove).


kanej (approved):
Looks good. I tested on windows.

Lets hold off on merging until we have the agnostic refactor cleared from dev.

PR #285, opened 2022-11-23
fix: avoid directories ending in .sol

antico5:


kanej:
What are the reproduction steps for the bug this resolves?

kanej (approved):
I was able to confirm the error doesn't appear in local testing.

PR #291, opened 2022-11-25
fix: files with whitespaces

### LINKED ISSUE #289: Whitespace issue with filenames not receiving diagnostics on windows
A case was found where contracts in a hardhat project in windows were not receiving diagnostics, even with correct config. It was found that the issue is having whitespaces on the path.

antico5:
This fixes an issue where any file with a whitespace on its path wouldn't get validation, regardless of OS.

Closes #289


kanej (approved):
LGTM!

PR #298, opened 2022-11-28
fix: set worker exit code to info and check worker status on watchedFilesChanged

antico5:
Worker exit code is not logged on an error line anymore. Additionally, worker status is checked when responding to onWatchedFilesChanged to avoid sending InvalidateCacheMessage to a dead worker.


kanej (approved):
Looks good.

I think we need to consider how to bring the rules in this file under test so we can document with regression tests the choices we are making.

PR #299, opened 2022-11-29
optimization: cache solc compiler paths

antico5:
This optimization was lost during the refactor so it's being brought back. It avoids checking/checksumming the compiler on each validation job.

### DIFF
--- a/server/src/services/validation/validate.ts
+++ b/server/src/services/validation/validate.ts
@@ -95,7 +95,10 @@ export async function validate(
 
       // Use bundled hardhat to compile
       await logger.trackTime("Compiling", async () => {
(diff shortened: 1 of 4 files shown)


server/src/services/validation/validate.ts
```diff
@@ -95,7 +95,10 @@ export async function validate(
 
       // Use bundled hardhat to compile
       await logger.trackTime("Compiling", async () => {
-        compilerOutput = await CompilationService.compile(compilationDetails!);
+        compilerOutput = await CompilationService.compile(
+          serverState,
+          compilationDetails!
```
kanej, line 100: Booooo!

kanej (approved):
LGTM!

PR #303, opened 2022-11-30
refactor: move server indexing and analysis to `initialize` step

### LINKED ISSUE #286: Move initialization to onInitialize step
Right now we are indexing, analyzing, spawning workers, etc, on the `onInitialized` handler. By this point in the lifecycle, the language client is able to send other requests, file events, etc, and the server is not yet on a state ready to serve them. Because of this, we use some state to determine if the server is ready, and ignore the incoming requests meanwhile. Also we need to artificially trigger file events (e.g. `didChangeContent`) after the initialization is over to get validation on opened files. 

Moving the initialization to the `onInitialize` step should fix this and we could get rid of the workarounds.

antico5:
Closes #286 

After this refactor, the lifecycle of the client/server is more according to the language specification. No requests can be sent from the client while initialization is not finished, so we don't have to check fox indexing to be finished on validation requests, and we don't have to send artificial didChangeContent requests.


client/src/extension.ts
```diff
@@ -40,7 +40,7 @@ export async function activate(context: ExtensionContext) {
   warnOnOtherSolidityExtensions(extensionState);
 
   return {
-    isReady: () => extensionState?.indexingFinished,
+    isReady: () => !!extensionState?.client?.initializeResult,
```
kanej, line 43: Do we know if this is necessary?

I can't find a usage in our codebase, was it added for tests then became unhooked? Or do we think that node language server needs it?

client/src/setup/setupIndexingHooks.ts
```diff
@@ -15,34 +13,14 @@ export function setupIndexingHooks(
...
-
-          indexingStatusItem?.dispose();
-
-          triggerValidationForOpenDoc(client);
-        }
-      );
+      // Hide spinner when initialization ends
+      indexingStatusItem.dispose();
 
       const fileIndexedDisposable = client.onNotification(
```
kanej, line 25: What do you think would be involved in getting rid of this?

client/src/setup/setupIndexingHooks.ts
```diff
@@ -104,38 +80,3 @@ function setupIndexingLanguageStatusItem(): LanguageStatusItem {
...
-  });
-}
-
-/**
- * Sends a no-op change notification to the server, this allows the
- * triggering of validation.
- * @param client the language client
- * @param textDoc the open text file to trigger validation on
- */
-function notifyOfNoopChange(client: LanguageClient, textDoc: TextDocument) {
```
kanej, line 124: When I test locally, we get diagnostics on the open editor when indexing closes. Given that we are not sending a noop change, is this what is happening?:

* lsp starts, initialization begins
* editor window with solc error is opened
* document open/change events are queued until initialization ends
* initializtion end, and the events trigger are played, and validation is run

Or is the mechanism something different?

client/src/types.ts
```diff
@@ -15,7 +15,6 @@ export interface ExtensionState {
   serverModulePath: string;
 
   client: LanguageClient | null;
-  currentIndexingJobs: LanguageStatusItem[];
```
kanej, line 18: Oh that is nice.

kanej (commented):
This is a good improvement. I love it when needless complexity can be stripped away.

What mental model should I have for how we deal with new workspace folders being added. It is just that actions that require an analysed file will do it on the fly, and the only thing that needs indexing to be finished is import completion, so they give no results until the indexing is complete. Is that right?

client/src/extension.ts
antico5: This is for the e2e tests, to know when the LSP finished initialization. Probably there are better ways of doing it but for now it's doing the job

client/src/setup/setupIndexingHooks.ts
antico5: Exactly, after the initialize request ends (server sends back a result), other requests can be sent to the server. At that point the client sends `textDocument/didOpen`

antico5:
> This is a good improvement. I love it when needless complexity can be stripped away.
> 
> What mental model should I have for how we deal with new workspace folders being added. It is just that actions that require an analysed file will do it on the fly, and the only thing that needs indexing to be finished is import completion, so they give no results until the indexing is complete. Is that right?

Yes! we are on the same page with that

client/src/extension.ts
kanej: Cool. That makes sense.

client/src/setup/setupIndexingHooks.ts
kanej: Is this the best way of approaching this though? I don't mean for this PR.

I am just interested in whether there is a better way of getting the language status item set per file. Is it possible just to set it on the open editor event through a custom call to the language server?

kanej (approved):

PR #307, opened 2022-12-05
Protocol tests

antico5:
Custom language client to perform protocol-level tests


server/src/services/validation/analyse.ts
```diff
@@ -16,6 +16,11 @@ export async function analyse(
     const solFileEntry = getOrInitialiseSolFileEntry(serverState, internalUri);
 
     await analyzeSolFile(serverState, solFileEntry, changeDoc.getText());
+
+    // Notify that a file was successfully analyzed
+    serverState.connection.sendNotification("custom/analyzed", {
```
kanej, line 21: Is this to support testing?

server/src/services/initialization/indexWorkspaceFolders.ts
```diff
@@ -22,10 +22,6 @@ export async function indexWorkspaceFolders(
   const logger = _.clone(serverState.logger);
   logger.tag = "indexing";
 
-  if (workspaceFolders.some((wf) => wf.uri.includes("\\"))) {
```
kanej, line 25: Happy to see the check gone. Is it the case now that the workspace folders uri are not converted over to unix style?

kanej (commented):
This looks great.

Can I ask how the lifetime of the client + language server works across the tests? Is it booted once rooted at the base of the three test projects? If I am writing a new test suite do I need to make sure docs and diagnostics (if I used them) are cleared?

Also, not for just now, while the tests are fresh in your mind is there anywhere you think we are lacking on the protocol test front (quickfixes say)?

antico5:
> This looks great.
> 
> Can I ask how the lifetime of the client + language server works across the tests? Is it booted once rooted at the base of the three test projects? If I am writing a new test suite do I need to make sure docs and diagnostics (if I used them) are cleared?
> 
> Also, not for just now, while the tests are fresh in your mind is there anywhere you think we are lacking on the protocol test front (quickfixes say)?

For the lifecycle, it's up to the test scenario on how to handle it. For example on initialize.test.ts, a brand new instance of the client (and then server) is created for an isolated folder. But for all the other tests, an instance is shared among them so it's only initialized once. It loads a folder with multiple projects inside it (like monorepo) and runs all the tests against them. The file opening/closing, cleanup of diagnostics, etc, is up to the user on the test scenarios. This gives maximum flexibility on how the client is used. So for instance multiple clients/servers could be instantiated on different folders to test different conditions, although for practicity having a shared one for general testing is better because initialization takes a couple seconds.

For the coverage, I migrated all of the existing tests that were on the old integration suite, but didn't add new ones. It could be helpful to assess the whole project and find where coverage could be missing. Still I think the current coverage counting e2e, protocol and unit tests is pretty good.

server/src/services/validation/analyse.ts
antico5: Yes. When a file is opened, a notification is sent to the server but it doesn't wait for a response, so this is in place to let the client know that we finished analyzing  an open file. Still I think in the future we should delegate this to the server, and have it wait on each of the requests that require a file to be analyzed. I.e. on a definition request, if the file is not analyzed yet, have that request wait for it instead of rejecting it. This would remove the need for this custom notification.

server/src/services/initialization/indexWorkspaceFolders.ts
antico5: As part of this PR there were some places where the `convertToUnixStyle` was removed. Eventually I'd like to get rid of that across the codebase and use uris/paths being agnostic of their representation. I think the helper method is no longer needed if we use the `vscode-uri` package

server/src/services/documents/onDidChangeContent.ts
```diff
@@ -30,12 +30,12 @@ export function onDidChangeContent(serverState: ServerState) {
     analyse: {
       action: analyse,
       changeActions: {},
-      wait: 240,
+      wait: process.env.NODE_ENV === "test" ? 0 : 240,
```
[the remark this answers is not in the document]
antico5: I thought of maybe having a config var for those values but for now I went for something simple. I'm not sure that we want to modify those values besides when testing. Also I don't think we should let the users tweak it either, I'd leave them as a sensible default.

server/src/services/validation/analyse.ts
kanej: Can we suppress this if the env is not testing then?

I really like the trigger analysis if it is not already started approach. I will add a linear ticket for that.

server/src/services/initialization/indexWorkspaceFolders.ts
kanej: I am happy with that. I think the main push in that direction originally was trying to get consistent results for the test suite.

If we are happy we are covered by the protocol test suite around uri then this is a good move.

server/src/services/documents/onDidChangeContent.ts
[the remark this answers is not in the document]
kanej: Okay. I'll buy that. I think if we see `process.env.NODE_ENC` proliferating then we should pull up these settings as config, but not expose them (as you suggest). Instead we have one place where we do `process.env.NODE_ENV` hackery.

test/protocol/test/textDocument/rename/foundry/rename.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  })
+
+  describe('[single-file][identifier] - rename', function () {
+    it('should rename', async () => {
+      const documentPath = getProjectPath('foundry/src/rename/Test.sol')
+      const documentUri = toUri(documentPath)
+
+      client.openDocument(documentPath)
+
+      const workspaceEdit = await client.rename(documentUri, makePosition(21, 17), 'newName')
```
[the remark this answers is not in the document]
antico5: Exactly, IMO asserting the workspaceEdit contents are enough for testing purposes, as applying it is on the clients responsibilities

server/src/services/validation/analyse.ts
antico5: Sure, will push the change now!

test/protocol/test/textDocument/rename/foundry/rename.test.ts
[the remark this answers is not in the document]
kanej: Utterly agree. I just really like that our test suites don't have to mutate the on disk files (and then clean up); that is beautiful.

kanej (approved):

PR #308, opened 2022-12-07
fix: foundry root imports

antico5:
This change fixes imports on foundry projects where they use paths relative to project root (i.e. `import 'src/MyContract.sol'`). It changes both the import resolution and the solc input as well. For foundry projects, `solcInput.sources` keys are not absolute paths, but paths relative to project root.  (i.e. `'/home/user/myProject/src/Contract.sol' => 'src/Contract.sol'`). This is the only way I found for json-interfaced solc to deal with circular dependencies that use root-relative imports.

### DIFF
--- a/server/src/utils/directoryContains.ts
+++ b/server/src/utils/directoryContains.ts
@@ -1,10 +1,13 @@
 import path from "path";
 
+/**
+ * Returns true if testPath is equal to dirPath or a subdirectory of it
+ * @param dirPath
+ * @param testPath
+ * @returns
+ */
 export function directoryContains(dirPath: string, testPath: string): boolean {
   const relative = path.relative(dirPath, testPath);
-  return (
-    !!relative &&
-    !relative.startsWith(`..${path.sep}`) &&
-    !path.isAbsolute(relative)
-  );
+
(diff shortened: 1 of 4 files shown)


server/src/utils/directoryContains.ts
```diff
@@ -1,10 +1,13 @@
...
+ */
 export function directoryContains(dirPath: string, testPath: string): boolean {
   const relative = path.relative(dirPath, testPath);
-  return (
-    !!relative &&
-    !relative.startsWith(`..${path.sep}`) &&
-    !path.isAbsolute(relative)
-  );
+
+  return !relative.startsWith(`..${path.sep}`) && !path.isAbsolute(relative);
```
kanej, line 12: Is there any danger here from os separator differences? Path relative is going to return os specific separators.

kanej (changes requested):
Can we put tests in to cover this change in logic?

server/src/utils/directoryContains.ts
antico5: You are right. That's why `path.sep` is used, so it works across OSs.

kanej: Ah, I missed that. Cool

kanej (approved):

PR #309, opened 2022-12-08
maintenance: update hardhat version to latest

### LINKED ISSUE #305: Validation blocked - Cannot read properties of undefined (reading '.js')
I create a new template project with command: npx hardhat and select typescript version.

When I open the .sol file with vscode, it has this problem: Validation blocked Cannot read properties of undefined (reading '.js')

My Create Procedure:

npm init -y
npx hardhat
npx hardhat compile

My environment:

OS Version: macOS. 12.6.1 (Apple Silicon M1)
Hardhat Version: 2.12.3
VSCode Version:  1.73.1
Hardhat Solidity Plugin Version: 0.6.3

antico5:
Closes #305 

By updating the LSP's bundled hardhat to the latest version, we no longer get the related error.

### DIFF
--- a/scripts/bundle.js
+++ b/scripts/bundle.js
@@ -62,7 +62,12 @@ async function main() {
     minifyWhitespace: true,
     minifyIdentifiers: false,
     minifySyntax: true,
-    external: ["vscode", "@nomicfoundation/solidity-analyzer", "fsevents"],
+    external: [
+      "vscode",
+      "@nomicfoundation/solidity-analyzer",
+      "fsevents",
+      "mocha",
+    ],
     platform: "node",
     outdir: ".",
     logLevel: "info",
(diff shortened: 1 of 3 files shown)


kanej (commented):
Should we bump the version in the integration projects as well?

My mental model is that we have no specific reason for having a different hardhat version in any of the hardhat integration project samples right now, so the simplest thing is to keep them in-sync with the main project until we do have an example.

antico5:
> 

I'll update the hardhat version of the test projects on the new protocol-tests branch, because I renamed the test/integration directory, so if I change it here there'll be a conflict when rebasing. Also just updated hardhat version to be 2.12.4

kanej (approved):
LGTM!

Issue #300, opened 2022-11-29
Move formatting to server side


antico5:
Right now we are executing code formatting (prettier/forge fmt) on the client side. We could move that to the server side by implementing the `textDocument/willSaveWaitUntil` method, or the `textDocument/formatting` method (probably VSCode issues the latter when the "format on save" setting is enabled)

kanej:
`vim.coc` supports the format command. So if we add it to the language server, vim gets it as well.

PR #316, opened 2022-12-20
fix: forge fmt respects project's foundry.toml

mds1:
Analogous to https://github.com/juanfranblanco/vscode-solidity/pull/363 (h/t @shekhirin), this PR ensures the formatter respects the `fmt` settings in the `foundry.toml` file


kanej:
Thanks @mds1, that is great. We will get it reviewed in the next few days.

client/src/formatter/forgeFormatter.ts
```diff
@@ -12,13 +12,22 @@ export async function formatDocument(
   );
 
   const formatted = await new Promise<string>((resolve, reject) => {
-    const forge = cp.execFile("forge", ["fmt", "--raw", "-"], (err, stdout) => {
-      if (err !== null) {
-        return reject(err);
-      }
+    const currentDocument = vscode.window.activeTextEditor?.document.uri;
+    const rootPath = currentDocument
+      ? vscode.workspace.getWorkspaceFolder(currentDocument)?.uri.fsPath
```
kanej, line 17: If I am following this right the move is to change passing the open editor version of the doc into `forge format` in the commandline, but instead of the previous behavior of a general invocation, we invoke in the root of the workspace, which likely has the `foundry.toml` file. In so doing we get the remapping.txt and the foundry.toml remappings considered in the format?

@antico5 assumedly we would want to run it in the root of the foundry project (to support monorepo setups). Is that information we have access to on the client side?

antico5: We don't have the project information on the client side. I agree with @kanej , it's not safe to assume that the workspace folder is the root of the project. Options I see for this are a) scan for project on client side (would avoid), b) ask for project to the server via custom request and c) move formatting to server side (mid-long term plan)

kanej: Is an approach here to take the change as proposed, as it is strictly an improvement in all cases, even if it is not covering the monorepo case. Then we open a task to move formatting from client side into the server itself where the foundry project information is available - and at that point support monorepo?

Or would you prefer we enhance this PR with the custom request approach?
@antico5

antico5: I'd accept this change and have an enhancement task for later, since for now there are higher priority tasks. This change is simple enough and effective in many cases in my opinion.

kanej: I'll update https://github.com/NomicFoundation/hardhat-vscode/issues/300 with that decision.

mds1: Just want to confirm my understanding that the plan is to merge this PR as-is, and improvements will happen in a follow up PR at some TBD point in the future? If you do need anything else from me here, let me know!

mds1:
Huh looks like some CI tests failed for `yarn run test:protocol`, but they do pass when locally

antico5:
@mds1 don't worry about that, macOS CI sometimes fails to download and install foundry and then some tests fail. This change is not breaking any tests

client/src/formatter/forgeFormatter.ts
kanej: @mds1 that is correct.

We will support the monorepo case in a separate PR when we move refactoring from the client into the language server (where we have more info).

We don't require any further changes.

kanej (approved):

PR #325, opened 2022-12-23
fix: foundry validation broken on windows

### LINKED ISSUE #324: Foundry validation not working on windows
After #308, the source locations handed to solc changed and in windows, because of backslashes, it's failing to match those locations with file URIs to correctly publish diagnostics

antico5:
Closes #324

### DIFF
--- a/server/src/services/validation/DiagnosticConverter.ts
+++ b/server/src/services/validation/DiagnosticConverter.ts
@@ -4,6 +4,7 @@ import { compilerDiagnostics } from "@compilerDiagnostics/compilerDiagnostics";
 import { CompilerDiagnostic } from "@compilerDiagnostics/types";
 import { Logger } from "@utils/Logger";
 import { SolcError } from "../../types";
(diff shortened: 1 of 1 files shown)


kanej (commented):
Can we add a language server integration test for foundry for this?

antico5:
> Can we add a language server integration test for foundry for this?

We have a couple tests that involve foundry validation, but they are skipped when running on windows. This is because on the CI, we are only installing foundry on macos and linux. The windows installation is trickier and is not present on the ci setup scripts. We could have a task for supporting foundry tests on windows on CI

kanej:
> > Can we add a language server integration test for foundry for this?
> 
> We have a couple tests that involve foundry validation, but they are skipped when running on windows. This is because on the CI, we are only installing foundry on macos and linux. The windows installation is trickier and is not present on the ci setup scripts. We could have a task for supporting foundry tests on windows on CI

Ah yes I remember now. Cool.

kanej (approved):

PR #327, opened 2022-12-23
chore: add new abi.encodeCall to completions

### LINKED ISSUE #319: bug: `abi.encodeCall` is not shown as a suggestion
Using v0.6.5 of the extension with Solidity 0.8.17 in a foundry project, and `abi.encodeCall` is not shown as an autocomplete suggestion, even though it's valid and I can use it in the code

image

antico5:
closes #319

### DIFF
--- a/server/test/services/completion/completion.ts
+++ b/server/test/services/completion/completion.ts
@@ -160,8 +160,8 @@ describe("Parser", () => {
           [
             "decode",
             "encode",
+            "encodeCall",
             "encodePacked",
-            "encodeWithSelector",
             "encodeWithSelector",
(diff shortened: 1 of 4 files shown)


server/test/services/completion/completion.ts
```diff
@@ -160,9 +160,9 @@ describe("Parser", () => {
           [
             "decode",
             "encode",
+            "encodeCall",
             "encodePacked",
             "encodeWithSelector",
-            "encodeWithSelector",
```
kanej, line 165: 🤦

kanej (approved):
LGTM!

PR #328, opened 2022-12-27
feat: function natspec completion

### LINKED ISSUE #296: Generate comments automatically for a function in solidity
Hi

I would like to suggest you support the auto-comment generation feature for solidity extension.

For example:
let's assume we have this function.

```
function setAdmin(address _admin, bool _enabled) external onlyOwner {
        _admins[_admin] = _enabled;
        
        emit SetAdminAccess(true);
    }
```
Then I want to have a comment skeleton like this when I write `/** ` and press `enter`.

```
/**
     * @notice
     *
     * @param _admin
     * @param _enabled
     */
```

I think this is possible because you are providing this for javascript/typescript already.

Looking forward to hearing from you.

Best regards,
Meanking

antico5:
Closes #296

When typing `/**`, which gets autocompleted to `/** */`, a Natspec completion item is shown to generate comments for the next found function declaration in the code.

natspec-completion

### DIFF
--- a/server/src/services/completion/onCompletion.ts
+++ b/server/src/services/completion/onCompletion.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
 import {
   VSCodePosition,
   CompletionList,
@@ -15,8 +16,9 @@ import {
   TextDocument,
   MemberAccessNode,
 } from "@common/types";
+import * as parser from "@solidity-parser/parser";
+import { FunctionDefinition } from "@solidity-parser/parser/src/ast-types";
 import { getParserPositionFromVSCodePosition } from "@common/utils";
-import { Logger } from "@utils/Logger";
 import { isImportDirectiveNode } from "@analyzer/utils/typeGuards";
 import {
   CompletionContext,
@@ -67,7 +69,8 @@ export const onCompletion = (serverState: ServerState) => {
           params.position,
           params.context,
           projCtx,
-          logger
+          serverState,
+          document
         );
 
         return { status: "ok", result: completions };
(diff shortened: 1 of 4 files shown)


EXTENSION.md
```diff
@@ -50,6 +50,10 @@ Relative imports pull their suggestions from the file system based on the curren
 
 ![Import completions](https://raw.githubusercontent.com/NomicFoundation/hardhat-vscode/main/docs/gifs/import-completion.gif "Import completions")
 
+Natspec documentation completion is also supported
+
+![Natspec completions](https://raw.githubusercontent.com/NomicFoundation/hardhat-vscode/main/docs/gifs/natspec-completion.gif "Natspec completion")
```
kanej, line 55: 👍

server/src/services/completion/onCompletion.ts
```diff
@@ -549,3 +559,71 @@ const getFullName = (node: Node | undefined): string | undefined => {
...
+    return null;
+  }
+
+  // Autocomplete with @dev, @param and @return tags
+  const range = {
+    start: position,
+    end: position,
+  };
+
+  let text = `\n * @dev Function description\n`;
```
kanej, line 607: I am not sure about including the filler text. I think it will be helpful the first few times then an annoyance.

server/src/services/completion/onCompletion.ts
```diff
@@ -549,3 +559,71 @@ const getFullName = (node: Node | undefined): string | undefined => {
...
+
+  for (const param of nextFunction.parameters) {
+    text += ` * @param ${param.name} parameter description\n`;
+  }
+
+  for (const _param of nextFunction.returnParameters ?? []) {
+    text += ` * @return return value description\n`;
```
kanej, line 614: Should be `@return ${param.name} value description;`

kanej (commented):
Is it possible to finish with the cursor inside `@dev <cursor>` on the comment?

kanej:
What do we think about including `@notice`? Maybe only for public functions?

server/src/services/completion/onCompletion.ts
```diff
@@ -549,3 +559,71 @@ const getFullName = (node: Node | undefined): string | undefined => {
...
+  }
+
+  return {
```
kanej, line 617: Based on the spec I think we can return snippet text, to allow for jumping by tab between sections:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion

image

server/src/services/completion/onCompletion.ts
antico5: Right! forgot about named returns

server/src/services/completion/onCompletion.ts
antico5: I will try to add this now

server/src/services/completion/onCompletion.ts
antico5: Sounds fair, I'll try to replace it with jump tabs

antico5:
> What do we think about including `@notice`? Maybe only for public functions?

I checked just a couple codebases and they seem to use `@dev` more. But I'm ok to include `@notice` for public functions. It's a bit awkward since I don't know how many non-devs are reading smart contracts :sweat_smile:

kanej (approved):
This is great. Lets hold off on merging it until the next release (with other natspec prs).

antico5:
Closing this one since #359 includes these commits

PR #333, opened 2022-12-30
chore: track timing for analysis

### LINKED ISSUE #332: Track timing on file analysis
Right now we are tracking the timing of file validation and most of the requests (completion, definition ,etc), but we are not tracking analysis.

I'd add a call to `trackTiming` on `analyse.ts` service.

antico5:
Closes #332


kanej (approved):
Good call. LGTM.

PR #359, opened 2023-01-04
Natspec completions

### LINKED ISSUE #296: Generate comments automatically for a function in solidity

### LINKED ISSUE #342: Support natspec in contract/interface/library completions
Similar to #328, provide a subset of natspec auto-inserts on a `/** */` completion before a contract or library or interface.

Based on the docs: https://docs.soliditylang.org/en/latest/natspec-format.html#tags, it looks like only `@title`, `@author`, `@notice` and `@dev` are supported. Lets include `@title` and `@dev` by default leaving `@notice` and `@author` out.

### LINKED ISSUE #343: Support natspec in event/state variable/public state variable
Similar to https://github.com/NomicFoundation/hardhat-vscode/pull/328, provide a subset of natspec auto-inserts on a /** */ completion before an event, state variable or public state variable.

This should be based on the docs: https://docs.soliditylang.org/en/latest/natspec-format.html#tags. 

For events it seems to `@dev` and `@param`.

But we should investigate why public state variables have an `@return` while state variables don't.

antico5:
closes #342
closes #343 
closes #296

This one is branched off a rebased version of PR #328


kanej (commented):
This is great.

I have pinged the channel to confirm people are happy with the suggestion we are making to users.

Reading through the spec I suspect we need to support completions on `///` comments as well. Does that require a different approach?

antico5:
> This is great.
> 
> I have pinged the channel to confirm people are happy with the suggestion we are making to users.
> 
> Reading through the spec I suspect we need to support completions on `///` comments as well. Does that require a different approach?

I think we can achieve that with some small adaptation. Should I add that to this PR?

kanej:
> > This is great.
> > I have pinged the channel to confirm people are happy with the suggestion we are making to users.
> > Reading through the spec I suspect we need to support completions on `///` comments as well. Does that require a different approach?
> 
> I think we can achieve that with some small adaptation. Should I add that to this PR?

Yes, lets add that to the requirement.

antico5:
@kanej triple forward slash format is implemented! ready to re-review

antico5:
@kanej ready for review!

kanej (approved):
I have tested and generated docs from the defaults using solc: https://docs.soliditylang.org/en/latest/natspec-format.html#documentation-output.

I am approving, but can I request one change which is that public state variables should not default to @dev, just @notice.

PR #360, opened 2023-01-05
fix: use uri instead of path for clearing diagnostics on valid compilation

### LINKED ISSUE #355: URI must contain a scheme
I know this isn't yet fully supported but I'm trying to run the server directly with Neovim's built in LSP client.

I'm seeing this error that seems to prevent the server from working at all:

```
URI must contain a scheme: /home/[...]/AccessControl.sol
```

This seems to be triggered by:

```
[TRACE][2023-01-03 17:34:03] .../lua/vim/lsp.lua:1045	"notification"	"textDocument/publishDiagnostics"	{  diagnostics = {},  uri = "/home/[...]/AccessControl.sol"}
```

Neovim seems to expect that the `uri` field will start with `file://` or some other scheme.

antico5:
Closes #355

### DIFF
--- a/server/src/services/validation/validate.ts
+++ b/server/src/services/validation/validate.ts
@@ -324,7 +324,7 @@ function validationPass(
     // TODO: improve this. Currently necessary because on hardhat source names are not full paths
     const docPath = openDocuments
       .map((doc) => doc.uri)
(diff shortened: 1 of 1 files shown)


kanej (approved):
LGTM!

PR #366, opened 2023-01-10
fix: improve forge binary lookup

### LINKED ISSUE #354: Assumes Foundry installation location
The location of the `forge` executable is assumed to be `~/.foundry/bin/forge` on non-Windows. I think this is the location where `foundryup` installs, but it can also be installed by other means in other locations in the `$PATH`.

```
Error: Command failed: ~/.foundry/bin/forge config --json
/bin/sh: line 1: ~/.foundry/bin/forge: No such file or directory
```

antico5:
Closes #354

### DIFF
--- a/client/src/formatter/forgeFormatter.ts
+++ b/client/src/formatter/forgeFormatter.ts
@@ -1,5 +1,7 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import * as vscode from "vscode";
 import * as cp from "child_process";
+import { runCmd, runningOnWindows } from "../utils/os";
 
 export async function formatDocument(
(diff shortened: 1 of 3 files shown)


kanej (commented):
This looks good.

I will do a round of tests on windows next.

kanej:
I have tested on windows by opening foundry and causing diagnostics to appear. I removed forge from my path so that the vscode terminal couldn't find `forge`, the extension used the fallback location instead.

kanej:
We should add the same fallback to the client for formatting. @antico5

antico5:
@kanej fallback added to client!

kanej (approved):

Issue #369, opened 2023-01-13
Does not work with the "src/=src/" remappings


PaulRBerg:
## Description

The `using for ... global` directive is explained in the [release announcement](https://blog.soliditylang.org/2022/03/16/solidity-0.8.13-release-announcement/) for Solidity v0.8.13, and my particular situation is explained [here](https://forum.soliditylang.org/t/by-allowing-the-global-using-for-directive-only-at-the-file-level-solidity-make-it-difficult-to-modularize-value-type-codebases/1410).

I have a bunch of imports like this:

```solidity
import "./Casting.sol" as C;
import "./Helpers.sol" as H;
import "./Math.sol" as M;
```

Which I am trying to apply to the `UD60x18` value type, like this:

```solidity
type UD60x18 is uint256;

using { C.intoSD1x18, C.intoUD2x18, C.intoSD59x18, C.intoUint128, C.intoUint40 } for UD60x18 global;
```
The Hardhat VSCode extension complains:

<img width="1006" alt="hardhat-using-for-global" src="https://user-images.githubusercontent.com/8782666/212305516-ab677b70-23c6-4d53-83a2-1e533d697138.png">

But I can happily compile my contracts with `forge build`.

## Source Code

See the [`src/ud60x18/ValueType.sol`](https://github.com/PaulRBerg/prb-math/blob/5085861101e592cde0ce6239031bb679cde7a2dc/src/ud60x18/ValueType.sol) file in PRBMath.

kanej:
Thanks for the report @PaulRBerg. I have been able to reproduce locally.

It looks like we are running the wrong solc version in this case for the diagnostics. We will have to investigate how that is happening.

antico5:
@PaulRBerg I found the issue, there's one line in the remappings.txt file that's making the extension misbehave in unsuspected ways, and it's the `src/=src/` line. 

Our current handling of remappings makes it so the solc input has the `remappings` setting to use absolute paths (we may need to change that), so the entry ends up being `src/=/absolute/path/prb-math/src/` on the solc input. When sent it that way, solc returns the errors that you see displayed, that don't seem related at all.

After removing the `src/=src/` line from remappings, our extension seems to work well with your project. Is this good enough of a solution for you? otherwise we may need to look into changing our remapping handling a bit on our side.

PaulRBerg:
Thanks a lot for digging this, @antico5.

[...]

PR #374, opened 2023-01-17
chore: upgrade @solidity-parser/parser version

antico5:
In order to support syntax from the latest versions of solidity.

related to issue #369


kanej:
I am getting type errors on the upgrade. I fear they have made quite a few changes!

PR #377, opened 2023-01-19
fix: remappings structure as solc input

### LINKED ISSUE #369: Does not work with the "src/=src/" remappings

antico5:
We are sending the remappings object to solc with absolute paths, e.g:
```
    "remappings": [
      "ds-test/=/home/.../projects/.../lib/forge-std/lib/ds-test/src/",
      "forge-std/=/home/.../projects/.../lib/forge-std/src/",
    ]
```

After this change, the remappings object is sent in the way they are declared, without transforming relative into absolute paths, e.g:
```
    "remappings": [
      "ds-test/=lib/forge-std/lib/ds-test/src/",
      "forge-std/=lib/forge-std/src/",
    ]
```

Closes #369

### DIFF
--- a/server/src/frameworks/Foundry/FoundryProject.ts
+++ b/server/src/frameworks/Foundry/FoundryProject.ts
@@ -102,8 +102,12 @@ export class FoundryProject extends Project {
     // Apply remappings to importPath if it's not a relative import
     if (!importPath.startsWith(".")) {
       for (const { from, to } of this.remappings) {
+        const toAbsolutePath = path.join(this.basePath, to);
         if (importPath.startsWith(from)) {
-          transformedPath = path.join(to, importPath.slice(from.length));
+          transformedPath = path.join(
+            toAbsolutePath,
+            importPath.slice(from.length)
+          );
         }
       }
     }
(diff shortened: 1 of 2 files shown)


antico5:
@kanej just checked and this one is good to go

kanej (approved):

PR #379, opened 2023-01-23
refactor: change state and notifications to have a single telemetryEnabled flag

### LINKED ISSUE #358: Combine `globalTelemetryEnabled` and `hardhatTelemetryEnabled` for language server
We should reduce the initialization flag and the update events to one `telemetryEnabled` flag within the language server. The extensions (vscode & vim.coc) should manage combining extension level settings down to one flag and keep the language server updated with that state.

antico5:
closes #358


kanej (commented):
I bumped the version of sentry and added a new guard against transactions when the telemetry flag is disabled. I was using transactions to test the flag and noticed they didn't turn off if the flag was disabled, this must have been an existing bug.

Could you review my changes. @antico5

antico5:
LGTM! nice catch with the transactions

PR #381, opened 2023-01-24
chore: convert to monorepo

### LINKED ISSUE #335: Convert to monorepo
We currently support three components:

* client - the extension
* language server - the standalone language server
* vim.coc - a vim coc extension

Along with the packages are our integration tests.

The packages relationships are managed with ad-hoc structures. We may add more packages in the future.

For both those reasons we should consider switching to a monorepo setup. It will hopefully allow us to automate more of the release process as well.

We would take `npm 7` workspaces as our baseline assumption for monorepo management.

antico5:
- Set up monorepo with following packages:
    - "client",
    - "server",
    - "coc",
    - "test/e2e",
    - "test/e2e/projects/main",
    - "test/protocol" 
- Remove `postinstall` script since the workspaces feature handles installing dependencies on packages
- Set `--ignore-platform` as default for install, since we need all the variants of `solidity-analyzer`
- Adjust `.vscodeignore` and bundling scripts accordingly with dependency hoisting

All tests pass, and from my manual testing it looks like the vsix is packed properly, and coc client/server are bundled correctly as well.

Note that when switching to this branch, or once this is merged, one should delete node modules and trigger a clean `yarn install` to avoid problems.

Closes #335


kanej:
I have pushed a commit that swaps from `yarn` to `npm`.

PR #382, opened 2023-01-24
chore: reduce server dependencies

### LINKED ISSUE #373: Reduce coc-solidity deps
The install time for `coc-solidity` is high (27s). We suspect this is npm pulling in all the deps.

We should consider switching the language server to a dev dependency, as we bundle it avoiding downloading it twice.

antico5:
This moves all dependencies (except `solidity-analyzer`) to devDependencies on the server package. This drastically reduces the installation time of the coc extension and the size from ~180mb to ~20mb.

Closes #373

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -63,10 +63,7 @@
     "rimraf": "3.0.2",
     "sinon": "12.0.1",
     "ts-node": "10.4.0",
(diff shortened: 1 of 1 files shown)


kanej (approved):
LGTM!

PR #383, opened 2023-01-25
optimization: only analyze local files on initialization

### LINKED ISSUE #302: Only analyze project local files on initialization stage
Right now we analyze all solidity files found under project folders. That includes libraries as well. A huge speed boost to the initialization stage could be analyzing just the local files (e.g. source paths for hardhat, src/test/script for foundry). 

A useful metric for tracking this is the `indexing` transaction p50-p95 on sentry

antico5:
Mark which indexed files are local (i.e. inside configured sources folder), and only analyze those during the initialization step. This avoids e.g. indexing all of OZ contracts when a small project uses few of them.

Closes #302

### DIFF
--- a/server/src/parser/common/types/index.ts
+++ b/server/src/parser/common/types/index.ts
@@ -398,6 +398,11 @@ export interface ISolFileEntry {
    * Has an analysis pass succeeded, allowing a user to operate on the ast.
    */
   isAnalyzed(): boolean;
+
+  /**
+   * Local means the file is part of the project's scope, and not external, dependency or library
+   */
+  isLocal: boolean;
 }
 
 /**
--- a/server/src/services/initialization/indexWorkspaceFolders.ts
+++ b/server/src/services/initialization/indexWorkspaceFolders.ts
@@ -106,10 +106,13 @@ export async function indexWorkspaceFolders(
     serverState.indexedWorkspaceFolders.push(workspaceFolder);
   }
 
-  // Analyze files
+  // Analyze local files
   await logger.trackTime("Analyzing solidity files", async () => {
     const span = sentryTransaction?.startChild({ op: "analyzeSolidityFiles" });
-    await analyzeSolFiles(serverState, logger, solFileUris);
+    const localSolFileUris = solFileUris.filter(
+      (uri) => serverState.solFileIndex[uri]?.isLocal === true
+    );
+    await analyzeSolFiles(serverState, logger, localSolFileUris);
     span?.finish();
   });
 }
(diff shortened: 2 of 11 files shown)


server/src/services/initialization/indexWorkspaceFolders.ts
```diff
@@ -162,27 +165,32 @@ export async function indexSolidityFiles(
       serverState,
       path.dirname(fileUri)
     );
+    let isLocal = true;
 
     for (const indexedProject of indexedProjects) {
```
kanej, line 170: I would refactor this lookup loop to a function.

server/src/services/initialization/indexWorkspaceFolders.ts
```diff
@@ -162,27 +165,32 @@ export async function indexSolidityFiles(
...
       }
     }
 
-    serverState.logger.trace(`Associating ${project.id()} to ${fileUri}`);
+    serverState.logger.trace(
+      `Associating ${project.id()} to ${fileUri}. Local: ${isLocal}`
+    );
 
     const docText = await serverState.workspaceFileRetriever.readFile(fileUri);
     serverState.solFileIndex[fileUri] = SolFileEntry.createLoadedEntry(
```
kanej, line 188: Should the `createLoadedEntry` take `isLocal` as a param?

server/src/parser/common/types/index.ts
```diff
@@ -398,6 +398,11 @@ export interface ISolFileEntry {
    * Has an analysis pass succeeded, allowing a user to operate on the ast.
    */
   isAnalyzed(): boolean;
+
+  /**
+   * Local means the file is part of the project's scope, and not external, dependency or library
+   */
+  isLocal: boolean;
```
kanej, line 405: I am not sure about `isLocal` as the name. It really indicates that a file should be primed and preanalysed during initialization, or is there more that we intend to do based on this property?

kanej (commented):
I have done a bit of testing on mac and ran into no problems across navigation and import completion. This is looking good.

I wonder whether, now that analysis and adding to the index for completions are separate, we should consider reading/analyzing the files for one project as a step and to have that happen after a project has initialized. So instead of all projects initializing then all files reading/analysing, we have a project initializes and then reading/analysing. Actually, lets skip on that idea; it might allow us to get to running for some projects more quickly (in a big repo), but the complexity will not be worth it - we will get complex bugs because it means indicating the lsp is ready for messages before initialization is complete for all projects.

server/src/parser/common/types/index.ts
kanej: Pato has confirmed that is how the concept is named in HH. So no need to change.

kanej (approved):
LGTM

PR #393, opened 2023-02-14
Truffle support

### LINKED ISSUE #45: Add Truffle adapter
Taken from feedback: https://github.com/nomiclabs/hardhat-vscode-feedback/issues/4

Currently, we only show compilation errors for Hardhat projects. Different tools can compile the same codebase differently, leading to different compilation errors. For example, they may choose to use a different version of solc for a certain source file.

We need to detect when a project is using Truffle and get the compilation errors based on its compilation pipeline, ideally without reimplementing its logic.

antico5:
- General
  - Identify projects that have truffle.js or truffle-config.js
  - Config file should be loadable (no errors) for validation to work. Same as hardhat
- Config
  - Solc version (semver format supported. i.e. `^0.8.0`
  - Contracts directory
- Import resolution
  - Absolute path
  - Relative path
  - Truffle direct imports (i.e. truffle/console.sol)
  - Local node modules contracts
  - Global node modules contracts

Closes #45


kanej:
I have been running through the truffle getting started guide:

https://trufflesuite.com/docs/truffle/quickstart/

After unboxing metacoin, I have a working project that I open with the extension:

```sh
truffle unbox metacoin [PATH/TO/DIRECTORY]
```

On opening the `./test/TestMetaCoin.sol` file I get red squiggly on direct truffle imports:

image

Specifically:

image

kanej:
> @kanej the `truffle/Assert.sol` import should be working. Can you verify that the contract is being identified as a truffle project from within the extension? also if there's anything wrong in the logs. The way it's resolved is by looking into either a local or global installation of truffle. For the `DeployedAddresses.sol` import, it won't work because it's a source generated dynamically by truffle. But truffle console, asserts, etc imports it should work.

What do you think of suppressing an error on the import of `truffle/DeployedAddresses.sol`?

antico5:
> > @kanej the `truffle/Assert.sol` import should be working. Can you verify that the contract is being identified as a truffle project from within the extension? also if there's anything wrong in the logs. The way it's resolved is by looking into either a local or global installation of truffle. For the `DeployedAddresses.sol` import, it won't work because it's a source generated dynamically by truffle. But truffle console, asserts, etc imports it should work.
> 
> What do you think of suppressing an error on the import of `truffle/DeployedAddresses.sol`?

I think it works. We can add more to that list if we discover there are other imports of this kind.

antico5:
> > @kanej the `truffle/Assert.sol` import should be working. Can you verify that the contract is being identified as a truffle project from within the extension? also if there's anything wrong in the logs. The way it's resolved is by looking into either a local or global installation of truffle. For the `DeployedAddresses.sol` import, it won't work because it's a source generated dynamically by truffle. But truffle console, asserts, etc imports it should work.
> 
> What do you think of suppressing an error on the import of `truffle/DeployedAddresses.sol`?

I tried this and actually found out that it doesn't quite work. I tried 2 options: 1st, suppressing the error, and 2nd, injecting a source with empty content for DeployedAddresses.sol. In the first case, all validation is lost because solc doesn't return other errors if there are import errors present, and in the 2nd case, the import line is not erroring but the usages e.g. `DeployedAddresses.MetaCoin()` will error because there's no such identifier.

server/src/frameworks/Truffle/TruffleProject.ts
```diff
@@ -0,0 +1,245 @@
...
+        fileSpecificErrors: {},
+        projectWideErrors: [
+          {
+            type: "general",
+            message: `Truffle project couldn't initialize correctly: ${this.initializeError}`,
+            source: "truffle",
+          },
+        ],
+      };
+      throw buildError;
```
kanej, line 167: Should this not originate within the initialize method?

antico5: The design I have in mind is as follows: on validation, we build the input to solc (framework- dependant), and then send it to solc. There are 3 types of errors: solc error, framework error on file (e.g. unresolvable import), and framework error on project (e.g. couldn't load config). The diagnostics and status item are handled differently depending on the type of error. For instance, framework project-level errors don't show as diagnostics but they show on the status item on files that belong to them.

During project initialization, if there's a failure e.g. loading the config, the error is stored in case compilation is attempted on a file belonging to such project.

kanej:
@antico5 is there an approach to resolving the truffle contracts under node modules as part of the truffle project?

The particular problem I am thinking of is from a truffle solidity test, jumping into the assert sol file via `import "truffle/Assert.sol";`, the file shows up as having project none, which I think leads to the imports not resolving.

server/src/frameworks/Projectless/ProjectlessProject.ts
```diff
@@ -31,7 +31,7 @@ export class ProjectlessProject extends Project {
   }
 
   public async fileBelongs(_file: string) {
-    return { belongs: true, isLocal: true };
```
kanej, line 34: Can you walk me through this?

Is this change that we don't precache analysis on unclaimed solidity files under a project where the config has errored?

server/src/frameworks/Truffle/TruffleProject.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  public id(): string {
+    return this.configPath;
+  }
+
+  public frameworkName(): string {
+    return "Truffle";
+  }
+
+  public async initialize(): Promise<void> {
```
kanej, line 53: Do we trigger a reinitialization on each config change?

antico5: Yes, that logic is on the `onWatchedFilesChanges` method of this class.

kanej: I pushed a change to get to the comparison check on `onWatchedFilesChanges` as the uri had `file://` as a prefix.

Is there a better way of doing that comparison with inbuilt vscode functions?

antico5: You could use `URI.parse(<uri>).fsPath`. The import is `import { URI } from "vscode-uri";`

antico5:
> @antico5 is there an approach to resolving the truffle contracts under node modules as part of the truffle project?
> 
> The particular problem I am thinking of is from a truffle solidity test, jumping into the assert sol file via `import "truffle/Assert.sol";`, the file shows up as having project none, which I think leads to the imports not resolving.
> 
> image

We could easily associate files under node_modules to the truffle project, although that wouldn't work if truffle is being used as a global package. For the global package, I don't know how we could associate them with a single project, since many open projects could be using the global module.

kanej:
> > @antico5 is there an approach to resolving the truffle contracts under node modules as part of the truffle project?
> > The particular problem I am thinking of is from a truffle solidity test, jumping into the assert sol file via `import "truffle/Assert.sol";`, the file shows up as having project none, which I think leads to the imports not resolving.
> > image
> 
> We could easily associate files under node_modules to the truffle project, although that wouldn't work if truffle is being used as a global package. For the global package, I don't know how we could associate them with a single project, since many open projects could be using the global module.

Lets take that approach. I think we can assume that a global install is less common and just deal with local `node_modules`.

antico5:
> > > @antico5 is there an approach to resolving the truffle contracts under node modules as part of the truffle project?
> > > The particular problem I am thinking of is from a truffle solidity test, jumping into the assert sol file via `import "truffle/Assert.sol";`, the file shows up as having project none, which I think leads to the imports not resolving.
> > > image
> > 
> > 
> > We could easily associate files under node_modules to the truffle project, although that wouldn't work if truffle is being used as a global package. For the global package, I don't know how we could associate them with a single project, since many open projects could be using the global module.
> 
> Lets take that approach. I think we can assume that a global install is less common and just deal with local `node_modules`.

Done!

kanej (approved):
I have tested on the latest getting started guide for truffle and we are providing a significantly improved experience.

We should hold off on merging until we have a release schedule (and consideration of docs)

PR #394, opened 2023-02-14
chore: docs for coc.nvim users

### LINKED ISSUE #384: Add instruction for vim users on how to check output logs
Since we don't have status items on vim to show project problems, like dependencies not installed or wrong config files, the only clue for finding the problem would be on the language server logs. It's not widely known how to access those, so we may add the instructions for vim users for:
- Checking the server logs by running `:CocCommand workspace.showOutput` -> solidity
- They should run `:CocRestart` after installing dependencies (e.g. openzeppelin contracts)

antico5:
Closes #384


kanej (approved):
LGTM!

PR #395, opened 2023-02-15
optimization: initialize projects sequentially

### LINKED ISSUE #389: throttle project initialization
We have reports of high resource usage in repositories with a large number of hardhat projects.

We currently initialize all projects in parallel. To lessen the peak load we should throttle the number of in-parallel project initializations. This will extend the time to get to a fully setup language server, but flatten the curve of cpu/memory usage.

It is an open question whether the throttling mechanism should be limited to hardhat projects or all projects (hh projects create a thread per project).

antico5:
Closes #389

### DIFF
--- a/server/src/services/initialization/indexWorkspaceFolders.ts
+++ b/server/src/services/initialization/indexWorkspaceFolders.ts
@@ -65,25 +65,23 @@ export async function indexWorkspaceFolders(
 
   // Append to global project map if they are not already indexed
   await logger.trackTime("Initializing projects", async () => {
-    await Promise.all(
-      foundProjects.map(async (foundProject) => {
-        if (foundProject.id() in serverState.projects) {
-          return;
-        }
+    for (const foundProject of foundProjects) {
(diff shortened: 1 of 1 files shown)


kanej (commented):
Do we have a reference repo with lots of projects? How much does sequential initialization add? Do we need to batch, say at 5 projects initializing at once?

antico5:
> 



> Do we have a reference repo with lots of projects? How much does sequential initialization add? Do we need to batch, say at 5 projects initializing at once?

What I did to test was opening a parent folder with a lot of child directories having projects in them. Each hardhat project takes a few seconds to initialize and non-hardhat projects take less than a second on average, because it's mostly just reading a file or executing a shell command.

Batching multiple projects' initialization is possible. I don't know what the sweet spot would be, between our old behavior (all parallel), and the new one (all sequential).

kanej:
We have made the decision to not implement batching now and to see how performance works in the wild based on our sentry analytics.

kanej (approved):

PR #399, opened 2023-02-17
Project error notifications

### LINKED ISSUE #390: Improve project level errors visibility
We leverage the status bar for showing project level errors, so a failure to validate the file due to an error in the related hardhat.config file for instance. We have reports of users not finding the error because the status bar is non-obvious (and a recent addition to vscode for this use case).

We opted for status bar over notifications because the vscode UX guidelines appeared to guide us to the status bar for workspace errors. If our users are unable to find the error, then this solution doesn't work.

We intend to try and improve the visibility of status bar errors by changing the background colour:

image

https://code.visualstudio.com/api/ux-guidelines/status-bar#error-and-warning-status-bar-items

If, in testing, this does not prove enough, we will switch to showing a short lived notification.

antico5:
This change will show a notification when validation is attempted on a file whose project is not correctly initialized. This is shown once per project at most, per session. It makes use of the `window/showMessage` notification from LSP, which makes it work on with all language clients besides vscode.

Closes #390


kanej (approved):
LGTM!

PR #421, opened 2023-03-08
fix: correctly assign project, analyze and validate on new/renamed file

### LINKED ISSUE #310: Clear diagnostics when a file is closed/renamed
Report from Omar:

I noticed that when I open a .sol file, the diagnostics pane is populated as expected, but when the file is closed/renamed, the errors persist.
There is no way to dismiss them. And in cases where the file is removed/renamed, my only solution is to physically reload VS Code to clear that list.Should we be clearing file diagnostics when its editor is closed? Thoughts?

### LINKED ISSUE #387: Direct imports not working with deeply nested folders
I have the following folder structure:

```
hardhat.config.ts
test
    folder1
        folder1.1
            Contract1.sol
        folder1.2
            Contract2.sol
    folder2
        Contract3.sol
        Contract4.sol
```

I'm doing `import "hardhat/console.sol";` in all of the solidity files, however they only work in `folder2`, and do not work in `folder1.1`, `folder1.2`.

I get the following error `Source "hardhat/console.sol" not found: File not found. Searched the following locations: "".solidity(6275)`

I also do `import 'openzeppelin-contracts-06/math/SafeMath.sol';` and get the error `Source "openzeppelin-contracts-06/math/SafeMath.sol" not found: File outside of allowed directories.solidity(6275)`

In my package.json 

```
{
    "devDependencies": {
        "openzeppelin-contracts-06": "npm:@openzeppelin/contracts@3.4.2",
    }
}
```

### LINKED ISSUE #416: Associate new solidity files with the correct project
On adding a solidity file to the project, associate it with the appropriate project metadata.

Part of #387

### LINKED ISSUE #417: Update internal file keeping on file move event
The language server's internal state maintains an index of each solidity file and the project it is related to. We need to update the metadata on a file rename event occurring in the project.

See the rename event in the specification: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didRename

Consideration should be given to dealing with existing diagnostics. The diagnostics should migrate with the rename.

Similarly we should update the metadata and diagnostics on file delete.

Part of #387

antico5:
I noticed the `didChange` event is fired before the `didWatchedFilesChange`, so files were being analysed/validated without their project being correctly assigned. I moved this assigning to the validate/analysis actions inline, so they are always up to date. Works with new files and renames.

Also included the "clear diagnostics on file delete" task.

Closes #417 
Closes #416 
Closes #387
Closes #310

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -39,7 +39,7 @@
     },
     "client": {
       "name": "hardhat-solidity",
-      "version": "0.6.11",
+      "version": "0.6.12",
       "license": "MIT",
       "dependencies": {
         "@ignored/solidity-language-server": "0.6.9",
@@ -206,10 +206,10 @@
     },
     "coc": {
       "name": "@ignored/coc-solidity",
-      "version": "0.6.11",
+      "version": "0.6.12",
       "license": "MIT",
       "dependencies": {
-        "@ignored/solidity-language-server": "0.6.11"
+        "@ignored/solidity-language-server": "0.6.12"
       },
       "devDependencies": {
         "coc.nvim": "^0.0.80",
@@ -13401,7 +13401,7 @@
     },
     "server": {
       "name": "@ignored/solidity-language-server",
-      "version": "0.6.11",
+      "version": "0.6.12",
       "license": "MIT",
       "dependencies": {
         "@nomicfoundation/solidity-analyzer": "0.1.1"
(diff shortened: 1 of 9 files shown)


kanej (commented):
Are there any tests we can get against the changes to indexing?

This looks great.

PR #422, opened 2023-03-08
fix: forge fmt on monorepos

### LINKED ISSUE #410: forge fmt not picking up settings from foundry.toml
As reported by `bassquake` on discord:

```
I'm having some problems with the formatter when set to "forge". I would expect this to read in the foundry.toml settings to pick up things like Line Length, but it seems like it still doesn't follow that rule.
```

Additionally they mention that it's a monorepo setup.

antico5:
Closes #410


client/src/formatter/forgeFormatter.ts
```diff
@@ -18,10 +20,8 @@ export async function formatDocument(
   const forgeCommand = await resolveForgeCommand();
 
   const formatted = await new Promise<string>((resolve, reject) => {
-    const currentDocument = vscode.window.activeTextEditor?.document.uri;
-    const rootPath = currentDocument
-      ? vscode.workspace.getWorkspaceFolder(currentDocument)?.uri.fsPath
-      : undefined;
+    const documentUri = vscode.window.activeTextEditor?.document.uri.fsPath;
+    const rootPath = documentUri && path.dirname(documentUri);
```
kanej, line 24: Can you walk me through this change. We are switching from running forge in the workspace folder (which fails if your a monorepo and the foundry project is a subfolder).

Instead we run forge fmt in the directory of the file to be formatted and forge does the work of finding the relevant `foundry.toml`?

kanej (approved):
I am not sure what is going on here, probable needs rebased, but the forge commit is good for merge!

PR #423, opened 2023-03-08
chore: dependabot for solidity-parser

### LINKED ISSUE #418: chore: use dependabot to track parser changes
We should update the parser when new releases become available, this can be partially automated by dependabot, see:

https://github.com/NomicFoundation/truffle-flattener/blob/master/.github/dependabot.yml

We want to add a similar config for the language server and other packages.

antico5:
Closes #418


kanej (approved):
LGTM!

PR #429, opened 2023-03-15
feat: server document formatting

### LINKED ISSUE #300: Move formatting to server side

antico5:
- Removed all formatting related code from client
- Added `textDocument/formatting` handling on the server
- Adapted/fixed the formatting logic to work on the server side properly
- Added extension config sync between client and server, since config is not naturally available from the server
- Tested and added some docs for coc-solidity
- Protocol tests for all formatting providers

Closes #300


client/src/setup/setupLanguageServerHooks.ts
```diff
@@ -92,12 +93,16 @@ const startLanguageServer = (extensionState: ExtensionState): void => {
     notifyTelemetryChanged
   );
 
-  const hardhatTelemetryChangeDisposable = workspace.onDidChangeConfiguration(
-    notifyTelemetryChanged
-  );
+  const configChangedDisposable = workspace.onDidChangeConfiguration(() => {
```
kanej, line 96: The idea here is to push the entire config to the language server? Rather than tracking individual config changes?

coc/README.md
```diff
@@ -16,6 +16,13 @@ In your vim/neovim, run this command:
 :CocInstall @ignored/coc-solidity
 ```
 
+## Configure
+
+In your coc-settings.json, the following settings are supported:
+
+- `"@ignored/coc-solidity.telemetry": true|false`
+- `"@ignored/coc-solidity.formatter": "prettier"|"forge"|"none"`
```
kanej, line 24: These will need renamed on a rebase.

client/src/setup/setupLanguageServerHooks.ts
antico5: Yes, the entire config is synced to the server

kanej: Can I confirm I have understood the code. The complete extension config is synced, but we retain the previous approach to telemetry enabled (which I think is fine - I just want to double check my mental model)?

kanej: In an out of band discussion, the telemetry update at the hardhat level has been subsumed within extension config updates, but the global setting still remains separate.

kanej (approved):
Looks good.

PR #441, opened 2023-03-27
chore: replace NODE_ENV to VSCODE_NODE_ENV to avoid clashes

### LINKED ISSUE #439: Use of NODE_ENV prone to clashing
As reported by Web3Uchiha on discord, they encountered the error `Cannot find module '/home/chirag/.vscode/extensions/node_modules/@ignored/solidity-language-server/out/index.js'`. It's because in our client we are using the `NODE_ENV` var to determine the server module path. We may consider changing the variable that we use, or use a different source of truth for determining we are in development mode.

antico5:
Closes #439


kanej (approved):
LGTM!

PR #442, opened 2023-03-28
chore: upgrade to GA 4

### LINKED ISSUE #392: Replace google analytics universal
Our current google analytics will stop processing on July 1st. We need to either convert to GA4 or find an alternative replacement.

We use GA to give us metrics on new and returning users. Cohort analysis is a nice to have, but we can live without it.

antico5:
These changes completely replace the reporting to our UA property with reporting to our new GA4 property. The endpoints and payloads change but the information we send is the same. 

On the Analytics dashboard, I created custom dimensions for extension version, language client and operating system. I also created a collection + report to show charts with these dimensions. It's under Reports -> Extension Analytics -> Users. 

GA takes around 24 hours to dump the latest information into the reports, so to see realtime events while testing, you can use the `Reports -> Realtime` feature.

Closes #392


server/src/analytics/GoogleAnalytics.ts
```diff
@@ -1,26 +1,21 @@
...
-  Analytics,
-  AnalyticsPayload,
-  DefaultRawAnalyticsPayload,
-  RawAnalyticsPayload,
-} from "./types";
+import { Analytics, AnalyticsPayload } from "./types";
 
-const GOOGLE_ANALYTICS_URL = "https://www.google-analytics.com/collect";
+const GA_URL = "https://www.google-analytics.com/mp/collect";
+const GA_SECRET = "5ddn8Rl5QwKk51Uls_BM6A";
```
kanej, line 8: We should consider how we inject this. I think we injected the GA app identifier from the top level index file; but maybe we should push further and have it loaded from a json file injected in the packaging process? Though we would probably need a switch for development.

The reason I bring it up, is I suspect we get some false positives on foundry because someone has forked the code but still kept the analytics ids the same.

server/src/analytics/GoogleAnalytics.ts
```diff
@@ -63,75 +55,39 @@ export class GoogleAnalytics implements Analytics {
...
+        extensionVersion: { value: this.extensionVersion },
+        languageClient: { value: this.clientName },
+        operatingSystem: { value: os.type() },
+      },
+      events: [
+        {
+          name: taskName,
+          params: {
+            engagement_time_msec: "10000",
+            session_id: machineId,
```
kanej, line 73: Shouldn't this be a randomly generated id on startup?

kanej (commented):
This is looking great.

What is involved in:

* the migration? Is it just that we release, or is there an export of data that we need to do?
* what is the testing process?

server/src/analytics/GoogleAnalytics.ts
antico5: Good point. You are suggesting something like reading a non-versioned secret file or an env variable at the time of packaging ?

server/src/analytics/GoogleAnalytics.ts
antico5: You are correct, forgot about this part.

server/src/analytics/GoogleAnalytics.ts
kanej: I am not sure. Maybe having a constants file pulled in at the top index with an empty set of default values. In development read the values from `env` variables, and during packaging write out the `env` variables to the constants file.

So anyone forking the codebase will get the empty version, but we can setup the environment variables for local development and release with no loss - we get analytics in production and in local development.

Maybe. Open to other approaches?

kanej:
Can we investigate migrating the active user history from the previous GA to our new GA property. Whether we succeed or not we should export our history user usage report.

server/src/analytics/GoogleAnalytics.ts
```diff
@@ -38,11 +39,7 @@ export class GoogleAnalytics implements Analytics {
 
   public async sendPageView(taskName: string): Promise<void> {
     try {
-      if (
-        this.serverState?.env !== "production" ||
```
kanej, line 48: Do we not want to block ga in dev?

antico5: I removed this condition while testing and then thought it may be ok to allow analytics on development, but I can roll back that change

server/src/analytics/GoogleAnalytics.ts
kanej: We have created a constants file that has empty default values, but optionally populates using env variables.

We therefore don't keep our analytics codes in the repo but can include them during the build.

PR #451, opened 2023-04-05
fix: suppress access errors while fast-globbing

### LINKED ISSUE #448: Error: EACCES: permission denied, scandir '/home/xxxx/Documents/xxxx/data/mongo/.m...
Our logs are clogged by:

Sentry Issue: [VSCODE-EXTENSION-W43](https://nomic-labs.sentry.io/issues/4051558030/?referrer=github_integration)

```
Error: EACCES: permission denied, scandir '/home/xxxx/Documents/xxx/data/mongo/.mongodb'
```

We should find our where we can get these sorts of file access errors, and determine their impact and then either solve them or suppress them.

antico5:
The error happened when there was a file or folder without access permissions while globbing for e.g. config files. It prevented any projects from being indexed at all.

Closes #448

### DIFF
--- a/server/src/utils/WorkspaceFileRetriever.ts
+++ b/server/src/utils/WorkspaceFileRetriever.ts
@@ -13,6 +13,7 @@ export class WorkspaceFileRetriever {
       cwd: baseUri,
       ignore,
       followSymbolicLinks: false,
+      suppressErrors: true,
     });
 
     return relativePaths.map((rp) =>


server/src/utils/WorkspaceFileRetriever.ts
```diff
@@ -13,6 +13,7 @@ export class WorkspaceFileRetriever {
       cwd: baseUri,
       ignore,
       followSymbolicLinks: false,
+      suppressErrors: true,
```
kanej, line 16: Beautiful.

kanej (approved):

PR #453, opened 2023-04-18
refactor: replace insertSemicolonCommand with additionalTextEdits

antico5:
As apart of the discarded work for command server execution, there was this refactor which is nice to have. Now, instead of relying on a command for testing and appending semicolon at the end of import lines, we make use of the `additionalTextEdits` field of completion items.


server/src/services/completion/getImportPathCompletion.ts
```diff
@@ -50,13 +57,32 @@ export function getImportPathCompletion(
...
+      line: position.line,
+      character: 0,
+    },
+    end: {
+      line: position.line,
+      character: 99999,
+    },
+  });
+
+  if (!/;\r?\n?$/.test(line ?? "")) {
```
kanej, line 74: Is this case covered by any tests?

antico5: I  think it's not. I'll add one now.

antico5: @kanej done!

kanej (approved):
LGTM!

PR #455, opened 2023-04-18
Fix: typos

omahs:
Fix: typos


kanej (approved):
This is great. Thanks for the fixes. Someday I will learn how many `d`s are in overridden.

PR #458, opened 2023-04-19
feat: Ape framework implementation

### LINKED ISSUE #391: Add Ape adapter
Add a development framework adapter for [Ape](https://github.com/ApeWorX/ape-solidity).

Ape is the successor to brownie, and is targeted at the python ecosystem.

Supporting Ape may get us most of the way to brownie support, so it is probably worth assessing how much further work is implied by brownie support.

The solidity support is in this repo: https://github.com/ApeWorX/ape-solidity

antico5:
- Index ape-config.yaml files for ape projects
- Read from config:
  - solidity version
  - contracts directory
  - dependencies directory
  - remappings
  - dependencies

- Installation of dependencies is checked and enforced before being operational

- Dependency support:
  - through version
  - through branch name
  - local dependencies

Closes #391


antico5:
@kanej marking this as WIP, as I noticed it needs some polishing. will get it ready for tomorrow

server/src/frameworks/Hardhat/HardhatProject.ts
```diff
@@ -47,7 +47,7 @@ export enum WorkerStatus {
 const REQUEST_TIMEOUT = 5000;
 
 export class HardhatProject extends Project {
-  public priority = 2;
+  public priority = 3;
```
kanej, line 50: Can you walk me through your current mental model of how priority works? Why is hardhat 3 for instance?

test/protocol/projects/ape/src/.cache/DappToolsERC20/dappnix/erc20.sol
```diff
@@ -0,0 +1,30 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.4.8;
+
+// Token standard API
+// https://github.com/ethereum/EIPs/issues/20
+
+contract ERC20 {
```
kanej, line 19: This is to test dependencies right? That is why a `.cache` file is being explicitly committed?

server/src/utils/WorkspaceFileRetriever.ts
```diff
@@ -14,6 +14,7 @@ export class WorkspaceFileRetriever {
       ignore,
       followSymbolicLinks: false,
       suppressErrors: true,
+      dot: true,
```
kanej, line 17: Are there any performance implications for including `.` files for our other uses?

server/src/frameworks/shared/buildBasicCompilation.ts
```diff
@@ -45,17 +48,22 @@ export async function buildBasicCompilation(
   }
 
   // Build solc input
-  const filePathsToCompile = _.map(dependencyDetails, "absolutePath");
   const sources: { [uri: string]: { content: string } } = {};
-  for (const filePath of filePathsToCompile) {
+  const remappings: string[] = [];
+
+  for (const { sourceName, absolutePath } of dependencyDetails) {
```
kanej, line 54: Can you walk me through what triggered this switch to absolute paths?

server/src/frameworks/Ape/ApeProject.ts
```diff
@@ -0,0 +1,233 @@
...
+
+    try {
+      const configContent = readFileSync(this.configPath).toString();
+      const config: ApeConfig = YAML.parse(configContent);
+
+      this.configSolcVersion = config.solidity?.version;
+
+      const configSourcesPath = (
+        config.contracts_folder ?? "contracts"
+      ).replace("~", homedir());
```
kanej, line 59: Is there a way of doing this with `path` manipulations?

server/src/frameworks/Ape/ApeProject.ts
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  public async initialize(): Promise<void> {
+    this.initializeError = undefined; // clear any potential error on restart
+
+    try {
+      const configContent = readFileSync(this.configPath).toString();
+      const config: ApeConfig = YAML.parse(configContent);
+
+      this.configSolcVersion = config.solidity?.version;
```
kanej, line 55: What happens if solidity isn't defined?

server/src/frameworks/Ape/ApeProject.ts
```diff
@@ -0,0 +1,233 @@
...
+
+      const configSourcesPath = (
+        config.contracts_folder ?? "contracts"
+      ).replace("~", homedir());
+
+      this.sourcesPath = path.resolve(this.basePath, configSourcesPath);
+
+      this.dependenciesPath = path.join(this.sourcesPath, ".cache");
```
kanej, line 63: The cache is in the `src` folder? Booooo.

server/src/frameworks/Ape/ApeProject.ts
```diff
@@ -0,0 +1,233 @@
...
+      }
+    }
+    return;
+  }
+
+  public getImportCompletions(
+    _position: Position,
+    _currentImport: string
+  ): CompletionItem[] {
+    return [];
```
kanej, line 197: What would be involved in getting import completions.

kanej:
I added some more APE protocol tests, mainly to remind myself how our protocol tests work.

client/src/setup/setupLanguageServerHooks.ts
```diff
@@ -48,8 +48,10 @@ const startLanguageServer = (extensionState: ExtensionState): void => {
         workspace.createFileSystemWatcher("**/hardhat.config.{ts,js}"),
         workspace.createFileSystemWatcher("**/{truffle-config,truffle}.js"),
         workspace.createFileSystemWatcher("**/foundry.toml"),
+        workspace.createFileSystemWatcher("**/ape-config.yaml"),
         workspace.createFileSystemWatcher("**/remappings.txt"),
         workspace.createFileSystemWatcher("**/*.sol"),
+        workspace.createFileSystemWatcher("**/.cache/*.sol"),
```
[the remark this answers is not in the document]
antico5: It's multiple directories deep. But I just noticed this line can be removed, since the previous one will trigger for sol files inside `.cache` as well.

coc/src/index.ts
```diff
@@ -28,6 +28,7 @@ export async function activate(context: ExtensionContext): Promise<void> {
           coc.workspace.createFileSystemWatcher(
             "**/{truffle-config,truffle}.js"
           ),
+          coc.workspace.createFileSystemWatcher("**/ape-config.yaml"),
```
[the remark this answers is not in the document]
antico5: Referencing my previous comment, I noticed that it's not needed, the generic `"**/*.sol"` watcher is enough.

server/src/frameworks/Ape/ApeIndexer.ts
```diff
@@ -0,0 +1,22 @@
+import path from "path";
+import { WorkspaceFolder } from "vscode-languageserver-protocol";
+import { decodeUriAndRemoveFilePrefix } from "../../utils";
+import { ProjectIndexer } from "../base/ProjectIndexer";
+import { ApeProject } from "./ApeProject";
+
+export class ApeIndexer extends ProjectIndexer {
+  public async index(folder: WorkspaceFolder) {
+    const uri = decodeUriAndRemoveFilePrefix(folder.uri);
+    const configFiles = await this.fileRetriever.findFiles(
```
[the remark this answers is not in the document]
antico5: It could happen that a dependency is an ape project on itself (has an ape config file), and we want to avoid indexing those as projects. It's the same approach we currently follow with all framework adapters.

server/src/frameworks/Hardhat/HardhatProject.ts
antico5: The adapter with the highest priority will have claim over a source file in case they are competing. I'm updating hardhat to priority = 4 (to avoid clashing with truffle, as ape was being developed in parallel). This would end up in the following order of priority: Hardhat > Truffle > Foundry > Ape > Projectless.

test/protocol/projects/ape/src/.cache/DappToolsERC20/dappnix/erc20.sol
antico5: That is correct

server/src/utils/WorkspaceFileRetriever.ts
antico5: I tested our indexing benchmark on our test monorepos and the performance is not impacted by enabling this flag.

server/src/frameworks/shared/buildBasicCompilation.ts
antico5: We always used absolute paths for building basic compilation (projectless, foundry). In the particular case of foundry, after building the basic compilation we transform the source keys to root-relative imports. 

Some extra logic was added to `getDependenciesAndPragmas`, and used by `buildBasicCompilation`, in which we append to remappings the resolved source names. E.g. if the framework adapter resolved `@source/name.sol` to `/absolute/path.sol`, then on solc input's sources the key would be `/absolute/path.sol` and there'll be a remapping entry `'@source/name.sol' => '/absolute/path.sol'`. This leverages the adapter's ability to resolve paths, for helping building the solc input. Notice that in the ape adapter, no further remappings parsing is needed for getting a correct compilation. All is leveraged from `buildBasicCompilation`.

server/src/frameworks/Ape/ApeProject.ts
antico5: I used this trick because `path.resolve` doesn't handle the tilde. And from [ape's configuration docs](https://docs.apeworx.io/ape/stable/userguides/config.html), under `Contracts Folder`, it seems to support absolute paths using tildes.

server/src/frameworks/Ape/ApeProject.ts
antico5: Then `configSolcVersion` would be undefined, and we would infer the compiler version from source files, like we do on projectless.

server/src/frameworks/Ape/ApeProject.ts
antico5: Yes, I also found that strage

server/src/frameworks/Ape/ApeProject.ts
antico5: Probably iterating over registered dependencies and scanning them for contracts. I left it for a next iteration like we did with foundry's adapter.

test/protocol/projects/ape/ape-config.yaml
```diff
@@ -0,0 +1,23 @@
...
+dependencies:
+  - name: OpenZeppelin
+    github: OpenZeppelin/openzeppelin-contracts
+    version: 4.4.2
+  - name: DappToolsERC20
+    github: dapphub/erc20
+    branch: dappnix
+    contracts_folder: src
+  - name: LocalOZ
+    local: /home/antico/webapps/solidity_projects/4_openzeppelin
```
kanej, line 14: This isn't installed right? Does it have any effect?

kanej: I anonymized it!

kanej (approved):
LGTM!

PR #463, opened 2023-04-25
fix: send unix-style paths to solc on truffle compilation

antico5:
Solc doesn't build  properly when sending source keys in  windows-style format (backslashes). Then we enforce forward slashes when building the input.

Also this PRs adds a check for whether truffle is installed or not, because some dependencies can't be resolved if it's not installed (e.g. console, assert).


server/src/frameworks/Truffle/TruffleProject.ts
```diff
@@ -60,6 +60,18 @@ export class TruffleProject extends Project {
 
     this.globalNodeModulesPath = execSync("npm root --quiet -g").toString();
 
+    try {
+      // Ensure truffle is installed either globally or locally
+      require.resolve("truffle", {
```
kanej, line 65: This check is breaking the truffle tests. Truffle isn't found locally or globally, so it throws.

kanej (approved):
LGTM!

PR #464, opened 2023-04-25
chore: fix  failing tests on mac CI build

### LINKED ISSUE #459: Fix failing Foundry tests on Mac CI Build
It is often the case that the mac github action CI build fails to download Foundry, the Foundry tests then fail.

Ideally we would fix this so that the Foundry download succeeds. But fallback strategies might include:

* check if the foundry download succeeded otherwise don't run the foundry tests
* cache the foundry download (this is likely of limited help)
* can foundry be added to the github action base image?

antico5:
This change involves checking whether foundry is installed when attempting to run protocol tests that involve forge. On mac and windows, if forge is not available, tests will be skipped. On linux they will always be run.

Closes #459

### DIFF
--- a/test/protocol/src/helpers.ts
+++ b/test/protocol/src/helpers.ts
@@ -1,4 +1,6 @@
+import { existsSync } from 'fs'
 import { URI } from 'vscode-uri'
+import os from 'os'
 
 export function toUri(path: string) {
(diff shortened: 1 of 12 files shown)


kanej (approved):
LGTM!

Issue #456, opened 2023-04-19
`forge fmt` does not always respect project's Foundry config


PaulRBerg:
I keep bumping into the bug that should have been fixed by https://github.com/NomicFoundation/hardhat-vscode/pull/316.

Hardhat VSCode does not always pick up my `[fmt]` configuration in `foundry.toml`:

```toml
[fmt]
  bracket_spacing = true
  int_types = "long"
  line_length = 120
  multiline_func_header = "all"
  number_underscore = "thousands"
  quote_style = "double"
  tab_width = 4
  wrap_comments = true
```

In particular, it does not consistently follow the `bracket_spacing` rule, e.g. this:

```solidity
import { IPRBProxy } from "../../src/interfaces/IPRBProxy.sol";
```

Turns into this:

```solidity
import {IPRBProxy} from "../../src/interfaces/IPRBProxy.sol";
```

The bug is transitory, though. When I open the file again and save it, the fmt rules are picked up correctly,

I cannot pinpoint the root cause, but this happens very often - basically every day. Might there be a race condition somewhere in Hardhat VSCode?

For context, these are my VSCode settings:

```json
{
  "[solidity]": {
    "editor.defaultFormatter": "NomicFoundation.hardhat-solidity"
  },
  "[toml]": {
    "editor.defaultFormatter": "tamasfe.even-better-toml"
  },
  "npm.exclude": "**/lib/**",
  "solidity.formatter": "forge"
}
```

This happens in all of my repos, but for a particular example, you can take a look at [PRBProxy](https://github.com/PaulRBerg/prb-proxy).

kanej:
@PaulRBerg thanks for feedback.

Does the "losing the connectiowith the foundry config", happen immediately on opening the project in vscode or is it that it works correctly initially but then there is some point during editing that it starts failing?

PaulRBerg:
The second scenario - works correctly initially, but it then drifts away.

antico5:
@PaulRBerg hey there! I've investigated this issue and can't find where or how this is happening. The formatting process is pretty straightforward. To help troubleshoot it, I've submitted a PR that adds a bit of logging to the formatting process, so when it happens you could set the logging to verbose and we may be able to spot the source of the problem.

The PR is #465

PaulRBerg:
Interestingly enough, this problem stopped happening on my end.

I will tentatively close the issue in the meantime, but if it happens again I will follow your guidance and set the logging to verbose 👍

PR #465, opened 2023-04-25
chore: add tracing to document formatting

antico5:
This will help us troubleshoot issues with formatting. Setting the log level to 'verbose' right now gives us some information like the formatting input and the output, but we can't tell which formatter was invoked.

This change adds tracing to show what formatter is used, what binary in the case of forge, and in which directory it's invoked.

References #456

### DIFF
--- a/server/src/services/formatting/onDocumentFormatting.ts
+++ b/server/src/services/formatting/onDocumentFormatting.ts
@@ -36,7 +36,7 @@ export function onDocumentFormatting(serverState: ServerState) {
           return null;
       }
     } catch (error) {
-      logger.error(`Error formatting document ${uri} with ${formatter}`);
+      logger.info(`Error formatting document ${uri} with ${formatter}`);
       return null;
     }
   };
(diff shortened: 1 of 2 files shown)


server/src/services/formatting/onDocumentFormatting.ts
```diff
@@ -32,9 +36,7 @@ export function onDocumentFormatting(serverState: ServerState) {
           return null;
       }
     } catch (error) {
-      serverState.logger.error(
-        `Error formatting document ${uri} with ${formatter}`
-      );
+      logger.error(`Error formatting document ${uri} with ${formatter}`);
```
kanej, line 39: Should we not log the error?

We get logging errors in Sentry, but I think we want no errors in Sentry for failures due to syntax highlighting.

Maybe we just return error for now, and see what that looks like, and I can make modifications in later releases?

antico5: I'll change this to info. I just realized today that the `logger.error` would trigger a sentry error.

kanej: Agreed.

PR #471, opened 2023-04-27
chore: support node 14

antico5:
Changed the target on esbuild to node14. By default it is `esnext`.

Manually ran protocol tests on node 14 and they are passing


kanej (approved):
LGTM!

Issue #462, opened 2023-04-25
Extension becomes unusably slow for bigger projects


0xPhaze:
I've noticed a few times that when working in bigger projects, the plugin would just be too slow and not respond. I'm currently working on a personal project of mine, and just recently it suddenly got very slow. After writing a few more tests, the plugin stopped working completely, most notably, because I could not run the formatter anymore. I'm using `forge` as a formatter and can run `forge fmt` in the console without a problem. Changing the formatter also did not help. I was able to use it again after deleting the extension and some lines of code, but it quickly broke down again. If I can trigger this consistently, I'll share the repo.

My specs are:
Mac M2 Max 96GB

0xPhaze:
Here is where it fails for me: 
https://github.com/0xPhaze/solnum/commit/c592785882507c29b580d1ea8783c7a99e1b6fa5

Basically before the commit, everything is fine, but as soon as I uncomment all that code, it begins getting very slow. From there it depends on how long (sometimes almost instantly, sometimes after messing around a lot; might depend on individual specs), but eventually it just freezes up entirely and I have to disable the extension.

0xPhaze:
It starts out quite fast for me as well, but then gets to this state.. This is what I see on the language server output when it freezes.

```
[Trace - 7:21:39 PM] Sending request 'textDocument/formatting - (21)'.
Params: {
    "textDocument": {
        "uri": "file:///Users/lain/git/solidity/solnum/test/M32x32.t.sol"
    },
    "options": {
        "tabSize": 4,
        "insertSpaces": true
    }
}
```

I also get this window loading forever until I cancel it.

<img width="451" alt="image" src="https://user-images.githubusercontent.com/103113487/234668174-62120429-8161-4e17-90e6-171d7083a794.png">

Though the language server still seems to be responding. This is the output when I hover. The above ^ window is still there.
```
[Trace - 7:23:35 PM] Sending request 'textDocument/hover - (22)'.
Params: {
    "textDocument": {
        "uri": "file:///Users/lain/git/solidity/solnum/test/M32x32.t.sol"
    },
    "position": {
        "line": 766,
        "character": 4
    }
}


[Trace - 7:23:35 PM] Sending notification '$/cancelRequest'.
Params: {
    "id": 22
}
```

Once I do actually cancel though, I get the following output:

```
[Trace - 7:25:23 PM] Sending notification '$/cancelRequest'.
Params: {
    "id": 21
}


[...]

PR #472, opened 2023-04-27
fix: make forge fmt call async and set a timeout

antico5:
This changes how `forge fmt` is invoked from the server on a document formatting request. I replace the use of `execSync` with `exec`, so the server can continue handling other requests while forge is running, and I also added a 2 seconds timeout for the process. If it times out, an error will be logged on the output channel.

References #462

### DIFF
--- a/server/src/services/formatting/forgeFormat.ts
+++ b/server/src/services/formatting/forgeFormat.ts
@@ -3,6 +3,7 @@ import path from "path";
 import { TextDocument } from "vscode-languageserver-textdocument";
 import { TextEdit } from "vscode-languageserver-types";
 import { URI } from "vscode-uri";
+import { promisify } from "util";
 import { resolveForgeCommand } from "../../frameworks/Foundry/resolveForgeCommand";
 import { Logger } from "../../utils/Logger";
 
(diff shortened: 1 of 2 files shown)


server/src/services/formatting/onDocumentFormatting.ts
```diff
@@ -33,7 +33,7 @@ export function onDocumentFormatting(serverState: ServerState) {
       }
     } catch (error) {
```
kanej, line 34: Can we differentiate between errors that are due to syntax problems and ignore them, and pass on other errors to sentry as an error.

server/src/services/formatting/forgeFormat.ts
```diff
@@ -3,23 +3,33 @@ import path from "path";
...
+    cwd,
+  });
+
+  // Write the file contents to the process
+  const { child } = execPromise;
+  child.stdin?.write(text);
+  child.stdin?.end();
+
+  // Set a time limit
+  setTimeout(() => {
```
kanej, line 25: Can we cancel this timeout after the `await execPromise`?

I don't think it is a memory leak, I just don't want kills on closed child processes.

server/src/services/formatting/onDocumentFormatting.ts
antico5: What use would we have for them ? possible cases would be that `forge` could not be resolved (not installed), or some config errors. This would be shown to the user on the logs. I thought sentry errors would be for unhandled situations (maybe I'm wrong in this understanding)

kanej (approved):
LGTM!

PR #474, opened 2023-04-28
chore: stop capturing initialize errors

antico5:
This is to address the new errors appearing on sentry related to project initialization. I just realized that our logger creates error entries when we call the `.error` method. I replaced existing logging related to project initialization from `error` calls to `info`, to avoid this.

### DIFF
--- a/server/src/services/initialization/indexWorkspaceFolders.ts
+++ b/server/src/services/initialization/indexWorkspaceFolders.ts
@@ -83,7 +83,7 @@ export async function indexWorkspaceFolders(
       try {
         await foundProject.initialize();
       } catch (error) {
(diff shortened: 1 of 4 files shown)


server/src/services/initialization/indexWorkspaceFolders.ts
```diff
@@ -83,7 +83,7 @@ export async function indexWorkspaceFolders(
       try {
         await foundProject.initialize();
       } catch (error) {
-        logger.error(`Error initializing ${foundProject.basePath}: ${error}`);
```
kanej, line 86: Lets log the error in its entirety. So `logger.error(error)`, if it starts clogging up sentry I will suppress it with a later patch.

kanej (changes requested):

PR #477, opened 2023-05-02
fix: support absolute path remappings on foundry projects

### LINKED ISSUE #476: Support `allow_paths` in foundry.toml
This issue was raised on discord: https://discord.com/channels/750408878008827925/951537942239797338/1101880603701420083

A cli run of `forge` gives no errors but it fails in the vscode extension with the following setup:

image

I suspect this is that the Foundry adapter is not reading and using `allow_paths`.

antico5:
Replaced `path.join` with `path.resolve` when parsing remappings to allow for absolute path remappings.

Closes #476


kanej (approved):
LGTM!

PR #493, opened 2023-05-18
fix: move npm root to try/catch block and make it async

### LINKED ISSUE #486: Catch exceptions when determine npm root in truffle
We are leaking unhandled exceptions during project initialization in truffle from the npm global lookup:

```ts
this.globalNodeModulesPath = execSync("npm root --quiet -g").toString();
```

https://github.com/NomicFoundation/hardhat-vscode/blob/9cfb24bb204bdab9b6f58f7598467205cc401887/server/src/frameworks/Truffle/TruffleProject.ts#LL61C3-L61C3

We should add a try/catch and log (but not to sentry) the issue. We should also switch from a sync call to an async call.

antico5:
Closes #486


kanej (approved):
LGTM!

PR #499, opened 2023-05-25
feat: documentSymbol - project outline

### LINKED ISSUE #491: Document Symbol - Outline provider
Use Slang parser to implement `textDocument/documentSymbol` - document outline.

antico5:
This PR is ready for review, but It's submitted over the `semantic_highlighting` branch because it's forked from that one. And this  will make it easier to see the diff.

Closes #491 

image


kanej:
In local testing (i.e. the quickfix.sol file in the smoke test suite) I noticed if there is special characters (I think they are multi-byte unicode) the names in the outline are cut short.

image

Is this a slang error or a conversion error?

antico5:
> In local testing (i.e. the quickfix.sol file in the smoke test suite) I noticed if there is special characters (I think they are multi-byte unicode) the names in the outline are cut short.
> 
> image
> 
> Is this a slang error or a conversion error?

Good catch. I even included utf8 characters in our test files (just with `ñ`, will try with more). We don't do the conversion ourselves anymore so I'll see what's happening

PR #523, opened 2023-10-24
Semantic highlight and document symbols on slang 0.10

### LINKED ISSUE #42: Add OUTLINE
Raised in feedback: https://github.com/nomiclabs/hardhat-vscode-feedback/issues/18

The request is to populate the outline section of the explorer sidebar:

image

I believe this requires implementing: `documentSymbolProvider` - https://code.visualstudio.com/api/language-extensions/programmatic-language-features#show-all-symbol-definitions-within-a-document

Supporting `documentSymbolProvider` would also get us navigation to symbol.

### LINKED ISSUE #217: Support advanced syntax highlighting

### LINKED ISSUE #491: Document Symbol - Outline provider

antico5:
This PR includes support for two new language features, semantic highlighting and document symbols (project outline).

It achieves so by making use of the latest version of slang (0.10.1)

These are some key points from this PR:
- I noticed that accessing nodes and properties from the CST through slang takes a lot of time, probably because it interfaces through napi. So I store some of this information on a SlangNodeWrapper object that keeps the data on the node process after reading it just once.
- Features are enabled only on supported platforms (os and architecture). Regardless, all user's platforms are now registered in sentry so we can have some insight on platforms slang should be targeting.
- The features have been rewritten using the new Cursor api from slang, versus the old custom walking method implemented on the extension side.
- Feature flags were implemented. The way they work is by fetching a json file in a given github repo/branch (currently https://raw.githubusercontent.com/antico5/hardhat-vscode/flags/flags.json). The numbers should be between 0 (0%) and 1 (100%). These flags are re-fetched on each server initialization. To determine if a given user should have the feature enabled, their machine id is hashed using MD5 and the last 2 bytes are taken. Then they are divided by 65535 and the percentile is used. This way, when a given flag is changed e.g. from 0.1 to 0.2, all users that have it previously enabled will continue to do so.
- The features are now error tolerant because of the new version of slang. 

Closes #217 
Closes #42 
Closes #491


package.json
```diff
@@ -17,7 +17,7 @@
     "url": "https://github.com/NomicFoundation/hardhat-vscode/issues"
   },
   "scripts": {
-    "postinstall": "npm install --no-save --ignore-scripts --force @nomicfoundation/solidity-analyzer-win32-ia32-msvc@0.1.1",
+    "postinstall": "npm install --no-save --ignore-scripts --force @nomicfoundation/solidity-analyzer-win32-ia32-msvc@0.1.1 @nomicfoundation/slang-win32-ia32-msvc@0.10.1",
```
OmarTawfik, line 20: not sure I understand why are we installing this here, rather than transitively through the child package deps?

package.json
```diff
@@ -17,7 +17,7 @@
     "url": "https://github.com/NomicFoundation/hardhat-vscode/issues"
   },
```
OmarTawfik, line 18: Does this PR close/replace #499 and #484?

package.json
```diff
@@ -17,7 +17,7 @@
...
   "scripts": {
```
OmarTawfik, line 19: I wonder if we need to add a changeset for this PR?

client/scripts/bundle.js
```diff
@@ -198,6 +199,17 @@ async function main() {
         "@nomicfoundation/solidity-analyzer-linux-x64-gnu": "0.1.1",
         "@nomicfoundation/solidity-analyzer-linux-x64-musl": "0.1.1",
         "@nomicfoundation/solidity-analyzer-freebsd-x64": "0.1.1",
+
```
OmarTawfik, line 202: Same question here, I wonder why are we packaging all of these, given that `vsce` can detect dependencies automatically through `package.json`?

server/src/parser/slangHelpers.ts
```diff
@@ -0,0 +1,45 @@
...
+import { RuleKind, TokenKind } from "@nomicfoundation/slang/kinds";
+import { TextRange } from "@nomicfoundation/slang/text_index";
+import _ from "lodash";
+import { TextDocument } from "vscode-languageserver-textdocument";
+import { Range } from "vscode-languageserver-types";
+import { getPlatform } from "../utils/operatingSystem";
+
+export type SlangNode = RuleNode | TokenNode;
+export type NodeKind = RuleKind | TokenKind;
+export type NodeCallback = (node: SlangNodeWrapper) => void;
```
OmarTawfik, line 11: looks like `NodeCallback` is not used anywhere. Is this intended?

package.json
antico5: Yes, this replaces both PRs

client/scripts/bundle.js
antico5: I think this is here so the .vsix file includes all platform-specific packages and then the same  vsix can be used across all platforms.

server/src/parser/slangHelpers.ts
antico5: Good catch! it was left over from before the rework

server/src/services/initialization/featureFlags.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export async function fetchFeatureFlags(
+  state: ServerState
+): Promise<FeatureFlags> {
+  state.logger.info("Fetching feature flags");
+
+  try {
+    return await got
+      .get(
+        "https://raw.githubusercontent.com/antico5/hardhat-vscode/flags/flags.json",
```
OmarTawfik, line 32: I suggest including that file in this PR, and changing the URL to:
Developers can always edit the URL locally for testing.

```suggestion
        "https://raw.githubusercontent.com/NomicFoundation/hardhat-vscode/development/flags.json",
```

server/src/services/initialization/onInitialize.ts
```diff
@@ -40,22 +44,49 @@ export const onInitialize = (serverState: ServerState) => {
       workspaceFolders,
     });
 
-    // fetch available solidity versions
-    await updateAvailableSolcVersions(serverState);
+    // fetch available solidity versions and feature flags
+    const [flags] = await Promise.all([
```
OmarTawfik, line 48: super nit suggestion, just for readability.

```suggestion
    const [flags, _] = await Promise.all([
```

server/src/services/initialization/onInitialize.ts
```diff
@@ -40,22 +44,49 @@ export const onInitialize = (serverState: ServerState) => {
...
+        await indexWorkspaceFolders(
+          serverState,
+          serverState.workspaceFileRetriever,
+          workspaceFolders,
+          transaction
+        );
+
+        return { status: "ok", result: null };
+      },
+      {
```
OmarTawfik, line 83: I wonder if we should log these metadata as part of the `indexing` event, or a slamg-specific one, as it is unrelated to indexing the workspace.

server/src/services/documentSymbol/onDocumentSymbol.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      const parseTree = parseOutput.parseTree;
+      span.finish();
+
+      if (parseTree === null) {
+        const strings = parseOutput.errors.map((e: any) =>
+          e.toErrorReport(uri, text, false)
+        );
+
+        throw new Error(`Slang parsing error: ${strings.join("")}`);
```
OmarTawfik, line 79: nit: since they are multiline strings each:

```suggestion
        throw new Error(`Slang parsing error:\n${strings.join("\n")}`);
```

server/src/services/documentSymbol/onDocumentSymbol.ts
```diff
@@ -0,0 +1,122 @@
...
+      );
+
+      const parseTree = parseOutput.parseTree;
+      span.finish();
+
+      if (parseTree === null) {
+        const strings = parseOutput.errors.map((e: any) =>
+          e.toErrorReport(uri, text, false)
+        );
+
```
OmarTawfik, line 78: My understanding is that this request is triggered on partial/invalid input as well, right?
For example, as the user is typing. Do we consider this a failure and throw an `Error` when it happens?

server/src/services/documentSymbol/SymbolVisitor.ts
```diff
@@ -0,0 +1,68 @@
...
+  public abstract ruleKind: RuleKind;
+  public abstract symbolKind: SymbolKind;
+  public abstract nameTokenKind: TokenKind;
+
+  constructor(
+    public document: TextDocument,
+    public symbolBuilder: SymbolTreeBuilder
+  ) {}
+
+  public onRuleNode(node: RuleNode, cursor: Cursor): void {
```
OmarTawfik, line 21: Not sure I understand why `SymbolVisitor` is an `abstract class`, given that all its children are just data pairs. Are we planning on having custom logic per visitor? Otherwise, I wonder if this can be simplified into a simple map/dictionary of `ruleKind -> (symbolKind, nameTokenKind)`.

server/src/services/documentSymbol/SymbolVisitor.ts
```diff
@@ -0,0 +1,68 @@
...
+    const range = slangToVSCodeRange(this.document, cursor.textRange);
+
+    let symbolName = "-";
+    let selectionRange = slangToVSCodeRange(this.document, cursor.textRange);
+
+    // Find identifier
+    const childCursor = cursor.spawn();
+
+    do {
+      const identifierNode: TokenNode | null = childCursor.findTokenWithKind([
```
OmarTawfik, line 31: suggestion, since this is doesn't have to be an identifier:
```suggestion
      const nameToken: TokenNode | null = childCursor.findTokenWithKind([
```

server/src/services/documentSymbol/SymbolVisitor.ts
```diff
@@ -0,0 +1,68 @@
...
+        _.last(childCursor.pathRuleNodes).kind === this.ruleKind
+      ) {
+        symbolName = identifierNode.text;
+        selectionRange = slangToVSCodeRange(
+          this.document,
+          childCursor.textRange
+        );
+        break;
+      }
+    } while (childCursor.goToNext());
```
OmarTawfik, line 46: Question: is there a valid case where the first node we find with `nameTokenKind` is not the intended one?

server/src/services/documentSymbol/SymbolTreeBuilder.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (this.currentPath.length === 0) {
+      this.symbols.push(symbol);
+    } else {
+      const lastSymbol = this.lastOpenSymbol();
+
+      if (!lastSymbol) {
+        throw new Error("Attempting to close a symbol but none is open");
+      }
+
+      lastSymbol.children?.push(symbol);
```
OmarTawfik, line 33: nit: can this ever be undefined?

```suggestion
      lastSymbol.children!.push(symbol);
```

server/src/services/documentSymbol/onDocumentSymbol.ts
```diff
@@ -0,0 +1,122 @@
...
+      }
+
+      const text = document.getText();
+
+      // Get the document's solidity version
+      let span = transaction.startChild({ op: "solidity-analyzer" });
+      const { versionPragmas } = analyze(text);
+      span.finish();
+
+      versionPragmas.push("<= 0.8.19"); // latest supported by slang
```
OmarTawfik, line 56: suggesgt moving this logic (Slang version selection/latest supported version) from both editor features to `slangHelpers.ts`

server/src/services/documentSymbol/onDocumentSymbol.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      const solcVersion =
+        semver.maxSatisfying(solcVersions, versionPragmas.join(" ")) ||
```
OmarTawfik, line 59: I wonder why are we using `solcVersions` here? Even if one exists, but not supported by Slang, `new Language(v)` will fail/throw.
Should we use `Language::supportedVersions()` instead?

test/protocol/test/textDocument/semanticTokens/full.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  after(async () => {
+    client.closeAllDocuments()
+  })
+
+  test('provides highlighting for types, keywords, functions, numbers and strings', async function () {
+    const semanticTokens = await client.getSemanticTokensFull(toUri(testPath))
+
+    expect(semanticTokens).to.deep.equal({
+      data: [
+        2, 0, 6, 0, 0, 0, 7, 8, 0, 0, 2, 0, 9, 0, 0, 0, 10, 13, 2, 0, 2, 0, 6, 0, 0, 0, 7, 10, 2, 0, 1, 4, 7, 0, 0, 1,
```
OmarTawfik, line 30: I suggest dividing this into groups of 5 (one for each token) for readability:

```suggestion
[
    [2, 0, 6, getTokenTypeIndex(SemanticTokenTypes.class), 0],
    [0, 7, 8. getTokenTypeIndex(SemanticTokenTypes.enum), 0],
    ...
].flat()
```

server/src/services/semanticHighlight/highlighters/StringHighlighter.ts
```diff
@@ -0,0 +1,25 @@
+import { SemanticTokenTypes } from "vscode-languageserver-protocol";
+import { TokenKind } from "@nomicfoundation/slang/kinds";
+import { NodeType } from "@nomicfoundation/slang/cst";
+import { HighlightVisitor } from "../HighlightVisitor";
+import { SlangNodeWrapper } from "../../../parser/slangHelpers";
+
+const stringKinds = new Set([
```
OmarTawfik, line 7: Maybe I'm missing something, but aren't these things highlighted by the syntactic (textmate) default highlighter? Why do we need to go through the LSP to define them?

- `KeywordHighlighter`
- `NumberHighlighter`
- `StringHighlighter`

server/src/services/semanticHighlight/onSemanticTokensFull.ts
```diff
@@ -0,0 +1,154 @@
...
+              status: "internal_error",
+              result: emptyResponse,
+            };
+          }
+
+          // Register visitors
+          const builder = new SemanticTokensBuilder(document);
+
+          const visitors = [
+            new CustomTypeHighlighter(document, builder),
```
OmarTawfik, line 94: I wonder if we need to consider these additional types:

- Enums
- User defined value types.
- Errors.
- Libraries.
- etc...

server/src/services/semanticHighlight/highlighters/FunctionDefinitionHighlighter.ts
```diff
@@ -0,0 +1,21 @@
+import { SemanticTokenTypes } from "vscode-languageserver-protocol";
+import { NodeType } from "@nomicfoundation/slang/cst";
+import { RuleKind, TokenKind } from "@nomicfoundation/slang/kinds";
+import { HighlightVisitor } from "../HighlightVisitor";
+import { SlangNodeWrapper } from "../../../parser/slangHelpers";
+
+// Highlights function definitions
+export class FunctionDefinitionHighlighter extends HighlightVisitor {
+  public tokenKinds = new Set([TokenKind.Identifier]);
```
OmarTawfik, line 9: Since most of the CST is identifiers, and very few are function/contract/struct/etc... definitions, I wonder if the cursor should be using the same logic as symbols, by searching by the parent kind instead. Thoughts?

OmarTawfik (changes requested):
Great work!
Left a few suggestions/questions. Thank you.

package.json
kanej: I have to say it is fuzzy. The git blame on that line reveals:

> To support the CI on testing an ugly hack is needed to force install the
windows ia32 version of the solidity analyzer. This is the architecture
that vscode extension host claims to run under, though the windows arch
according to node will be x64. This hack is in the root package.json as
a `postinstall` script.

I suspect that there is a clash between the architecture as exposed by node when run within the vscode extension host in CI tests and when the npm install is happening.

Do we need the slang dep added here? Not sure, if we remove it and the CI passes tests I think we are fine - though we should test the slang features on windows!

server/src/services/semanticHighlight/highlighters/StructDefinitionHighlighter.ts
```diff
@@ -0,0 +1,21 @@
...
+import { SlangNodeWrapper } from "../../../parser/slangHelpers";
+
+// Highlights struct definitions
+export class StructDefinitionHighlighter extends HighlightVisitor {
+  public tokenKinds = new Set([TokenKind.Identifier]);
+
+  public enter(nodeWrapper: SlangNodeWrapper): void {
+    const ancestors = nodeWrapper.pathRuleNodes;
+    if (
+      nodeWrapper.type === NodeType.Token &&
```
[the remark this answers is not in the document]
OmarTawfik: cc @AntonyBlakey for providing binding info from Slang.

package.json
antico5: I'll test if it works without adding this to the postinstall

server/src/services/documentSymbol/SymbolVisitor.ts
antico5: Previously there was custom logic for some of the child classes, but it was changed a lot when it was reworked to use the Cursor api. I'll refactor it

server/src/services/documentSymbol/SymbolVisitor.ts
```diff
@@ -0,0 +1,68 @@
...
+  public abstract symbolKind: SymbolKind;
+  public abstract nameTokenKind: TokenKind;
+
+  constructor(
+    public document: TextDocument,
+    public symbolBuilder: SymbolTreeBuilder
+  ) {}
+
+  public onRuleNode(node: RuleNode, cursor: Cursor): void {
+    const range = slangToVSCodeRange(this.document, cursor.textRange);
```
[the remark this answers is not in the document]
antico5: Not intended, it was left over from an older iteration. Will remove it as part of the refactor into dictionary

server/src/services/documentSymbol/onDocumentSymbol.ts
```diff
@@ -0,0 +1,122 @@
...
+          e.toErrorReport(uri, text, false)
+        );
+
+        throw new Error(`Slang parsing error: ${strings.join("")}`);
+      }
+
+      const builder = new SymbolTreeBuilder();
+
+      const visitors: SymbolVisitor[] = [
+        new StructDefinition(document, builder),
```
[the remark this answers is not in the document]
antico5: Will add support for YulFunctionDefinition. I don't think we need to support the unnamed function, it's been deprecated for a long time.

server/src/services/semanticHighlight/highlighters/StructDefinitionHighlighter.ts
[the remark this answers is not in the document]
antico5: It's highlighting the definition and usage for me:
image
image

server/src/services/semanticHighlight/onSemanticTokensFull.ts
antico5: Definitely! I don't know how I missed those. Maybe because I covered them on document symbols and then forgot about highlighting

server/src/services/semanticHighlight/highlighters/StringHighlighter.ts
OmarTawfik: I suggest so, as this means extra work in the LSP, and extra data sent over the wire to the client, without visible user benefits.

package.json
OmarTawfik: Looks like we don't use it for now.

client/scripts/bundle.js
OmarTawfik: Just wondering, is this intended, or just a limitation of the current build?
My understanding is that `vsce` now supports platform-specific extensions natively:

https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions

package.json
OmarTawfik: Thanks for explaining!

If we need it here, I suggest adding this info as a comment above. For example:

```json
{
  "_scripts_comment_": [
    "To support the CI on testing an ugly hack is needed to force install the",
    "windows ia32 version of the solidity analyzer. This is the architecture",
    "that vscode extension host claims to run under, though the windows arch",
    "according to node will be x64. This hack is in the root package.json as",
    "a postinstall script.",
  ],
  "scripts": {
    "postinstall": "npm install --no-save --ignore-scripts --force @nomicfoundation/solidity-analyzer-win32-ia32-msvc@0.1.1 @nomicfoundation/slang-win32-ia32-msvc@0.10.1"
  }
}
```

server/src/services/documentSymbol/SymbolVisitor.ts
OmarTawfik: I see. Thanks for explaining!

In that case, if we are checking `cursor.ancestors().length` above, and it turns out it is greater than 1, I suggest calling `cursor.goToParent()` instead of `cursor.goToNext()`, as it will be able to skip the entire subtree.

server/src/services/documentSymbol/onDocumentSymbol.ts
OmarTawfik: I might be missing something, but I looked at the Slang implementation, and looks like we do indeed always return a non-nullable node. when is `parseTree === undefined` in that case?

server/src/services/documentSymbol/onDocumentSymbol.ts
```diff
@@ -0,0 +1,122 @@
...
+
+      const text = document.getText();
+
+      // Get the document's solidity version
+      let span = transaction.startChild({ op: "solidity-analyzer" });
+      const { versionPragmas } = analyze(text);
+      span.finish();
+
+      versionPragmas.push("<= 0.8.19"); // latest supported by slang
+
```
OmarTawfik, line 57: I also suggest using `Language::supportedVersions()` and adding both the first and last versions in that array, instead of hard coding it:

```suggestion
      const versions = Language::supportedVersions();
      versionPragmas.push(`>= ${versions[0]} <= ${versions[versions.len - 1]}`);

```

server/src/services/initialization/onInitialize.ts
OmarTawfik: I see. Thanks for explaining.

test/protocol/projects/hardhat/contracts/semanticTokens/full/SemanticTokens.sol
```diff
@@ -0,0 +1,54 @@
...
+        testContract contractAsLocalVar;
+
+        TestInterface interfaceAsLocalVar;
+        TestStruct memory structAsLocalVar;
+        contractAsLocalVar;
+        contractAsFuncParam;
+        interfaceAsFuncParam;
+        interfaceAsLocalVar;
+        structAsFuncParam;
+        structAsLocalVar;
```
OmarTawfik, line 42: not sure I understand what these lines are testing. These are expression statements, and all not highlighted. Is that correct?

server/src/services/semanticHighlight/highlighters/StructDefinitionHighlighter.ts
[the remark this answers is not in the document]
OmarTawfik: Aha! I was thrown off by parameter names being highlighted the same as their types, but only in their declaration, not usage. Looks like it exists already in production though. I wonder if it can be fixed by us highlighting the parameter names?

image

client/scripts/bundle.js
kanej: A limitation of the current build.

I have opened an issue for the platform specific visx file approach: https://github.com/NomicFoundation/hardhat-vscode/issues/534

OmarTawfik: Thanks for confirming. If we can still publish the extension with the added size, it is not a blocker then.

kanej: I think we can, @antico5 can you check the difference in size?

.github/workflows/ci.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v3
         with:
-          node-version: "14.x"
+          node-version: "18.x"
```
OmarTawfik, line 27: #533 fixed the CI and failing tests. I suggest rebasing on latest `development` to be able to run the new tests in this PR on windows.

server/src/services/semanticHighlight/highlighters/StructDefinitionHighlighter.ts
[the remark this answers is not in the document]
antico5: You are right, it's from the textmate grammar. On the screenshot I sent, I had the grammar turned off so I could only see the effects of the semantic highlighting

server/src/services/semanticHighlight/highlighters/FunctionDefinitionHighlighter.ts
antico5: I just tested and after removing keyword highlighters (because they are in the textmate grammar), highlighting is very fast. It takes around ~6ms on bigger contracts (excluding parsing). So I don't think we need to change this for now

test/protocol/projects/hardhat/contracts/semanticTokens/full/SemanticTokens.sol
antico5: Yes, that's correct. They were there because at some point they were incorrectly highlighted, but they shouldn't and we are asserting that

.github/workflows/ci.yaml
antico5: Done, rebased on top of it! thanks

test/protocol/test/textDocument/semanticTokens/full.test.ts
antico5: I'm afraid this could be too time consuming for what it's worth. My thought process is, once we verify locally that all symbols are highlighted properly (by running the test client), we can dump the LSP response and put it into the assertion. Then if at any point the logic changes, the test will fail. To achieve what you suggest the easiest way I can think of is fabricating cst-like nodes on the test case for increased readability and leveraging SemanticTokensBuilder class for serializing, and then performing the assertion. But I'm not sure it's worth the effort

server/src/services/documentSymbol/SymbolVisitor.ts
antico5: Actually I think it's ok to leave this as an object hierarchy. It gives us flexibility in case we want to change the abstraction, push some methods down to subclasses, etc. If I refactor this into a dictionary and then need to leverage those oop features I'd have to convert it again to an oop model.

server/src/services/documentSymbol/onDocumentSymbol.ts
antico5: Just removed the check for null

package.json
antico5: Looks like tests are passing without the postinstall script :tada:

server/src/services/documentSymbol/SymbolVisitor.ts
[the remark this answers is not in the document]
antico5: Removed `node` variable

server/src/services/documentSymbol/onDocumentSymbol.ts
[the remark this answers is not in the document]
antico5: Added support for both

server/src/services/semanticHighlight/onSemanticTokensFull.ts
antico5: Added support for these

server/src/server.ts
```diff
@@ -19,6 +19,8 @@ import { Telemetry } from "./telemetry/types";
 import { attachDocumentHooks } from "./services/documents/attachDocumentHooks";
 import { availableVersions } from "./services/initialization/updateAvailableSolcVersions";
 import { onDocumentFormatting } from "./services/formatting/onDocumentFormatting";
+import { onSemanticTokensFull } from "./services/semanticHighlight/onSemanticTokensFull";
+import { onDocumentSymbol } from "./services/documentSymbol/onDocumentSymbol";
 
 export default function setupServer(
```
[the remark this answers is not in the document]
antico5: @OmarTawfik tried the node js flags approach but it didn't work. The `report-on-fatalerror` doesn't work, tested with a sample program triggering the panic from slang. 

About the extension logs, it could be possible (althought not sure if viable) to scan for extension logs and try to pattern-match anything related to panics. For testing this, I'd need if you could rebase the `simulate-panic` branch on top of `main` on slang. Because otherwise I can't build the LSP because of some changes in the modules in the latest versions

server/src/services/initialization/featureFlags.ts
OmarTawfik: I learned about `ServerState.env` today. Maybe we can use it here?

- If `state.env == "production"` we can use that URL.
- If `state.env == "development"` we can load that file from the local repo.

Just a suggestion.

test/protocol/projects/hardhat/contracts/semanticTokens/full/SemanticTokens.sol
OmarTawfik: I suggest adding comments to indicate which is which.

server/src/services/documentSymbol/SymbolVisitor.ts
OmarTawfik: Alright. Let's keep it for now. Thanks!

server/src/services/semanticHighlight/onSemanticTokensFull.ts
```diff
@@ -0,0 +1,159 @@
...
+              visitor.enter(nodeWrapper);
+            }
+
+            cursor.goToNext();
+          }
+          span.finish();
+
+          return { status: "ok", result: { data: builder.getTokenData() } };
+        } catch (error) {
+          logger.error(`Slang parsing error: ${error}`);
```
OmarTawfik, line 153: I suggest specializing this `try/catch` to just Slang parsing, or updating the error title:

```suggestion
          logger.error(`Semantic Highlighting Error: ${error}`);
```

server/src/services/documentSymbol/onDocumentSymbol.ts
OmarTawfik: Fixed the typing issue in Slang. Thanks for letting me know!

server/src/services/semanticHighlight/onSemanticTokensFull.ts
```diff
@@ -0,0 +1,159 @@
...
+        const { versionPragmas } = analyze(text);
+        span.finish();
+
+        const versions = Language.supportedVersions();
+        versionPragmas.push(
+          `>= ${versions[0]}`,
+          `<= ${versions[versions.length - 1]}`
+        );
+
+        const solcVersion = semver.maxSatisfying(
```
OmarTawfik, line 66: left over:

```suggestion
        const slangVersion = semver.maxSatisfying(
```

server/src/services/documentSymbol/onDocumentSymbol.ts
```diff
@@ -0,0 +1,140 @@
...
+      const { versionPragmas } = analyze(text);
+      span.finish();
+
+      const versions = Language.supportedVersions();
+      versionPragmas.push(
+        `>= ${versions[0]}`,
+        `<= ${versions[versions.length - 1]}`
+      );
+
+      const solcVersion = semver.maxSatisfying(
```
OmarTawfik, line 65: left over:

```suggestion
      const slangVersion = semver.maxSatisfying(
```

server/src/services/semanticHighlight/onSemanticTokensFull.ts
OmarTawfik: Also, I suggest reusing `versions` created above instead of fetching it two more times.

server/src/services/documentSymbol/onDocumentSymbol.ts
OmarTawfik: Also, I suggest reusing `versions` created above instead of fetching it two more times.

test/protocol/test/textDocument/semanticTokens/full.test.ts
OmarTawfik: I see. Sounds good. Thanks.

client/scripts/bundle.js
OmarTawfik: > Including all the platform packages (current): 17.5 Mb

I think that would work. I tested a few popular extensions in the marketplace and they are 30-50mb.
Thanks for checking!

server/src/server.ts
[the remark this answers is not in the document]
OmarTawfik: Sure. I just rebased it on top of the `v0.10.1` tag, which is the version you are using. Does that work, or do you want to use `main` instead? it has a few unreleased (yet) changes that shouldn't affect you.

server/src/services/documentSymbol/SymbolVisitor.ts
OmarTawfik: Thanks for reporting! I'm not sure I can reproduce on my end. Can you please send me a test/snippet?
I would be happy to fix it and send you a new build.

Not blocking though.

server/src/services/semanticHighlight/highlighters/FunctionDefinitionHighlighter.ts
OmarTawfik: Great!

server/src/services/semanticHighlight/highlighters/StructDefinitionHighlighter.ts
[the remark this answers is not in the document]
OmarTawfik: Thanks.

server/src/services/documentSymbol/SymbolVisitor.ts
OmarTawfik: Apologies for the confusion. Looks like we are missing an API on our end. I will review it and get back to you. In the meantime, I think checking `cursor.ancestors().length` should be enough.

Side note: I'm also adding `cursor.parent(): Node` in the next release, to allow you to do both operations in one call. But not blocking at all!

OmarTawfik: Edit: apologies for the confusion.

`goToParent()` would go up, but then continue to revisit that parent, then its children (including the current node) again. What I think we need to do is to `goToParent()` and then `goToNextSibling()`, so when the cursor continues, it skips the entire subtree of the parent (including its children, and the current node).

Is that the behavior you want?

OmarTawfik: Edit: Apologies for the confusion. No need to call `.goToParent()` since you actually want to continue inspecting the remaining children.

server/src/server.ts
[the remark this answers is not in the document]
OmarTawfik: I also did some debugging today, and I found out that it helps if you set env variable `RUST_BACKTRACE` to `FULL`. It should print the panic traces in the stderr of the server process.

flags.json
```diff
@@ -0,0 +1,9 @@
+{
+  "documentSymbol": {
```
OmarTawfik, line 2: I suggest starting with `0` in this PR, and leave turning it on for future PRs.

server/src/services/initialization/featureFlags.ts
antico5: When on development mode, we have features enabled, we do that on the `isFeatureEnabled` function.

server/src/services/documentSymbol/onDocumentSymbol.ts
OmarTawfik: Additionally, because of the two implementations, when we don't find a suitable version:

- In document symbols, we throw an error, failing the request.
- In semantic highlighting, we log an error, and return an empty response instead.

I agree that we should log this, in order to to be notified when it happens. But realistically speaking, that is not always actionable. For example, the user might be user a version that is too old (we don't plan on supporting anything before `0.4.11`) or too new (hasn't been released yet, or we are still working on supporting it). In that case, should it be still an error? should we just log the fact, and still make sure we return an empty response?

What do you think?

server/src/server.ts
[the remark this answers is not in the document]
OmarTawfik: Let's put this aside for now, until we move to Slang `0.11.0`

server/src/services/documentSymbol/onDocumentSymbol.ts
antico5: Sounds great on moving to slangHelpers and logging + empty response on document symbols

server/src/services/semanticHighlight/onSemanticTokensFull.ts
antico5: This has been addressed

server/src/services/documentSymbol/onDocumentSymbol.ts
OmarTawfik: I see semantic highlighting is still throwing errors instead of catching it? Is this intended?

OmarTawfik (approved):

Issue #452, opened 2023-04-09
Add support for region comments in Solidity


denyshorman:
Region comments are useful for organizing code and making it more readable by allowing developers to collapse and expand sections of code. This is especially useful for large files with many functions or blocks of code that are related but not necessarily sequential.

There are several use cases where region comments could be useful:

**Contract sections**: Solidity contracts can have multiple sections, such as state variables, constructors, functions, and modifiers. Using region comments to group these sections together would make it easier to navigate and understand the structure of a contract.

```solidity
contract MyContract {
  //#region State variables
  uint public myVar;
  uint private myPrivateVar;
  //#endregion

  //#region Constructors
  constructor(uint _myVar) public {
    myVar = _myVar;
  }
  //#endregion

  //#region Public functions
  function getMyVar() public view returns (uint) {
    return myVar;
  }
  //#endregion

  //#region Modifiers
  modifier onlyOwner() {
    require(msg.sender == owner, "Only owner can call this function.");
    _;
  }
  //#endregion
}
```

**Code blocks**: Solidity code can be complex, and it's often helpful to break it down into smaller, more manageable sections. Region comments could be used to group related code blocks together, such as loops, conditionals, or function calls.

```solidity
function myFunction() public {
  //#region Setup
  uint[] memory myArray = new uint[](10);
  for (uint i = 0; i < 10; i++) {
    myArray[i] = i;
  }
  //#endregion

  //#region Loop
  for (uint i = 0; i < myArray.length; i++) {
    if (myArray[i] % 2 == 0) {
      doSomething(myArray[i]);
    }
  }
  //#endregion
}
```

Regional comments are [supported](https://code.visualstudio.com/docs/editor/codebasics#_folding) in multiple languages in VS Code, and I believe it would be beneficial to have support for region comments in Solidity as well. Solidity code can often be complex, with multiple sections and code blocks, and region comments would make it easier for developers to navigate and understand the code. The ability to collapse and expand sections of code would improve code readability and reduce the cognitive load on developers. Additionally, many developers are already familiar with using region comments in other languages, so having support for them in Solidity would make the development experience more consistent and intuitive.

kanej:
This is interesting. We are interested in whether other users would find this useful, so leave a comment or a `+1`.

[...]

PR #526, opened 2023-11-03
feat: add support for region comments

scorpion9979:
- [x] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team.
- [ ] I didn't do anything of this.

---
## Overview

This PR introduces support for region comments within Solidity `.sol` files previously mentioned in https://github.com/NomicFoundation/hardhat-vscode/issues/452, addressing the issue where developers were seeking the ability to organize code into collapsible sections for improved readability and navigation. 

## Changes

- Modified the Solidity language configuration file to recognize the `//#region` and `//#endregion` patterns.

## Implementation Details

Region comments are implemented by modifying the Solidity language grammar to include start and end region markers. The pattern follows the typical convention of `//#region` and `//#endregion` which is familiar to developers from other programming languages.

```solidity
//#region ExampleRegion

// Your Solidity code here

//#endregion ExampleRegion
```

The language configuration now includes logic to parse these markers and translate them into foldable regions within VS Code.

## Screenshots

_Included are screenshots demonstrating the collapsible regions in action within a Solidity file._

image

\
image

## How to test this feature

1. Check out this PR and follow the instructions in [CONTRIBUTING.md#running-locally](https://github.com/NomicFoundation/hardhat-vscode/blob/development/CONTRIBUTING.md#running-locally) to run it locally.
2. Open a `.sol` file in VS Code.
3. Insert `//#region` and `//#endregion` comments around the code you wish to collapse.
4. Verify that you can collapse and expand the region as expected.

### DIFF
--- a/docs/publish-extension.md
+++ b/docs/publish-extension.md
@@ -25,6 +25,9 @@ release.
 
 9. Push the release branch and open a pull request using the new changelog entry as the PR description
 10. Generate a release candidate vsix file with `npm run package`, the vsix file should appear in the `./client` folder with the new version number
+
+> NOTE: ensure .env file is populated with GA and Sentry secrets before packaging (see `./env.example`)
+
 11. Manually run smoke tests on the new features across:
 
 - mac os x
--- a/client/language-configuration.json
+++ b/client/language-configuration.json
@@ -32,6 +32,12 @@
     ["[", "]"],
     ["(", ")"]
   ],
+  "folding": {
+		"markers": {
+			"start": "^\\s*//\\s*#?region\\b",
+			"end": "^\\s*//\\s*#?endregion\\b"
+		}
+	},
   "onEnterRules": [
     {
       // e.g. /** | */


pegahcarter:
@Xanewok can we get some support on this feat? It would be great to have for navigating large contracts and can prevent a screen-split if you're working on, for example, public functions and their errors or events.

Xanewok:
I'd be open to it; I thought this is a Microsoft/C#-ism but it seems to be ubiquitously supported in languages like Go (https://github.com/microsoft/vscode/pull/69899), Python (https://github.com/microsoft/vscode-python/issues/33) and C++, C#, TS etc.

Ideally:
- it should probably be supported out of box in the VSCode at this point with a way to opt-out for specific languages
- or one should use a dedicated extension for this https://marketplace.visualstudio.com/items?itemName=maptz.regionfolder

However, since this is such a small addition, I think it's more beneficial to have it rather than the other way around, especially if users will be migrating from the existing Juan Blanco's extension that supports it.

@kanej what do you think?

docs/publish-extension.md
```diff
@@ -25,6 +25,9 @@ release.
 
 9. Push the release branch and open a pull request using the new changelog entry as the PR description
 10. Generate a release candidate vsix file with `npm run package`, the vsix file should appear in the `./client` folder with the new version number
+
+> NOTE: ensure .env file is populated with GA and Sentry secrets before packaging (see `./env.example`)
```
Xanewok, line 29: Could we drop this change, please?

PR #531, opened 2023-11-27
fix: compare ast ranges in isNodeEqual

puuuuh:
- [x] Because this PR includes a **bug fix**, relevant tests have been included.

There can be multiple nodes without names, and they can't be added via addChild/addTypeNode


antico5:
@puuuuh sorry for the long delay. If still relevant, could you share a minimal solidity example to show what language action (definition, implementation etc) doesn't work, and your proposed change fixes?

PR #554, opened 2024-04-15
Use newest Slang 0.14.0

Xanewok:


Xanewok:
I'm not sure why Socket flags these as new/removed? This only changes the slang dependency in the lockfile. Shouldn't it be performing the analysis with the equivalent of `npm ci`/frozen lockfile?

OmarTawfik (approved):
LGTM on the Slang side.

kanej (approved):

PR #562, opened 2024-04-22
feat: Use the latest version support by Slang even if not resolved

Xanewok:
This is arguably more useful approach - we designed our parser to be error-tolerant and it makes more sense to offer slightly incomplete experience rather than just suddenly turning off all features just because the user is using a slightly newer version that may not get full coverage.

### DIFF
--- a/server/src/parser/slangHelpers.ts
+++ b/server/src/parser/slangHelpers.ts
@@ -6,6 +6,7 @@ import { TextDocument } from "vscode-languageserver-textdocument";
 import { Range } from "vscode-languageserver-types";
 import { Language } from "@nomicfoundation/slang/language";
 import semver from "semver";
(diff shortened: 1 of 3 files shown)


Xanewok:
I verified locally that the outline works even with `pragma solidity ^0.8.24;` even though Slang does not support it (0.8.22 is the latest).

kanej:
Can I check, if the user is on `0.8.24` for example, we would treat it as `0.8.22` and attempt a parse.

If the code being parsed includes no unrecognized features we continue as before: full outline and semantic highlighting.

If the code includes new features, then we still get a partial parse; so outline will appear but it will have missing sections around the new features.

Is that right?

@Xanewok

Xanewok:
@kanej yep, that's exactly right.

PR #563, opened 2024-04-22
release: 0.8.2

kanej:
### Added

- Run Semantic Highlighting and document symbols on later solc versions not yet support by the Slang parser (behind feature flag) ([562](https://github.com/NomicFoundation/hardhat-vscode/pull/562))


kanej:
@SocketSecurity ignore npm/keytar@7.9.0

This was published 2 years ago. We get it via vsce - which we should probably update.

PR #612, opened 2024-11-14
feat(#534): platform specific builds of vsix

### LINKED ISSUE #534: Create platform specific builds of the vsix
We currently bundle every version of the `@nomicfoundation/solidity-analyzer` for every platform, so that one vsix can be used across all platforms.

We should switch to platform specific vsix files. To reduce the size of installs. This will involve reworking the release and build processes.

For details on platform specific vsix files and publication see:
https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions

gordon-to:
Create platform specific vsix builds. closes #534.

- Updated .github/workflows/ci.yaml
  - update cache action to v4
  - add arch and platform to matrix
  - set TARGET env var to platform-arch
- Update client/package.json to use TARGET env var in `vsce package` command

based on https://github.com/microsoft/vscode-platform-specific-sample/blob/main/.github/workflows/ci.yml


antico5:
@gordon-to Apologize for the long delay. Could you explain why we need platform-specific vsix files? we always distributed a single package through the vscode marketplace that should work on all supported platforms.

antico5:
Thanks for the contribution. We are about to release a new version that uses Slang 0.19 (check [this PR](https://github.com/NomicFoundation/hardhat-vscode/pull/635)) , which uses wasm instead of native extensions, so the package size will be greatly reduced. We can revisit this PR after the upgrade

antico5:
I'm closing this PR since after our upgrade to Slang 1.0.0 we won't have platform-specific dependencies anymore

PR #624, opened 2025-01-13
chore/update-typescript-eslint-and-module-resolution

kanej:
`Slang@19` supports being used from commonjs (which the vscode extension is). However, to get access to the exported types we require updating the module resolution we are using to `node16` for typescript builds.
This module resolution option requires we update the version of typescript being used.
This PR:

1. Updates a cut of packages that need updating to work with `typescript@5.7.3`, including
  - the microsoft language server lib
  - lodash types
  - node types
  - esbuild
  - eslint and plugins
3. Updates typescript to version `typescript@5.7.3`
4. Applies some linting fixes and some naming convention eslint disable lines
5. **Turns off** the typescript libcheck


test/e2e/tsconfig.json
```diff
@@ -1,7 +1,9 @@
 {
   "compilerOptions": {
-    "target": "es2016" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
-    "module": "commonjs" /* Specify what module code is generated. */,
+    "target": "es2022" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
+    "lib": ["es2023"],
```
alcuadrado, line 4: Are these compatible with recent versions of vscode?

kanej: This is based on:

* VSCode is on Node 20
* The tsconfig for working with Node 20 suggests the target and lib selection above: https://www.npmjs.com/package/@tsconfig/node20

However, this was only tentative, it should all be reviewed once the all the parts allow a run through of Slang with WASM.

There is a question about whether we should target Node 18 rather than Node 20 so that we are keeping to the Hardhat node version support promise. This wouldn't affect the VSCode library but you can run the language server independently.

alcuadrado: I think that if it works with vscode is enough. We don't need to keep the same support, as vscode updates on a different cadence.

coc/package.json
```diff
@@ -32,9 +32,7 @@
     "@nomicfoundation/solidity-language-server": "0.8.8"
   },
   "devDependencies": {
-    "coc.nvim": "^0.0.80",
-    "esbuild": "^0.16.0",
-    "eslint": "^7.23.0"
+    "coc.nvim": "^0.0.80"
```
antico5, line 35: Won't our build/lint scripts fail if we remove those dev dependencies?

antico5:
@kanej Besides the 1 question I left, I'm seeing the "package" step on the CI is failing. I don't see any changes in the bundle script so it's interesting. Should I look into it ?

PR #312, opened 2022-12-14
feature: solidity survey notification

antico5:

PR #625, opened 2025-01-14
chore: add the solidity survery 2024 popup

galargh:
- [ ] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team.
- [x] I didn't do anything of this.

---

This is a follow-up to https://github.com/NomicFoundation/hardhat/pull/6133

It adds the solidity survey popup similarly to how it was done in https://github.com/NomicFoundation/hardhat-vscode/pull/312

### DIFF
--- a/client/src/popups/showSoliditySurveyPopup.ts
+++ b/client/src/popups/showSoliditySurveyPopup.ts
@@ -29,7 +29,7 @@ export async function showSoliditySurveyPopup({
 
   const lastSeen = context.globalState.get<number>(SURVEY_LAST_SEEN_KEY);
 
-  if (lastSeen !== undefined && now > lastSeen + SURVEY_COOLDOWN_PERIOD) {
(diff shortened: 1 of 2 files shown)


client/src/popups/showSoliditySurveyPopup.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const completed = context.globalState.get<boolean>(SURVEY_COMPLETED_KEY);
+
+  if (completed === true) {
+    return;
+  }
+
+  const lastSeen = context.globalState.get<number>(SURVEY_LAST_SEEN_KEY);
+
+  if (lastSeen !== undefined && now > lastSeen + SURVEY_COOLDOWN_PERIOD) {
```
kanej, line 32:
```suggestion
  if (lastSeen !== undefined && now < lastSeen + SURVEY_COOLDOWN_PERIOD) {
```
I think this should be inverted. If the time is before the last seen + cool off period.

kanej (approved):

PR #635, opened 2025-02-06
Upgrade to slang 1.0.0 and use prettier 3 on the codebase

### LINKED ISSUE #632: Update Semantic Highlighting and Outline to use Slang@19
`Slang@19` comes with bug fixes and runs in WASM, but implies some changes to the API.

Because `Slang@19` is ESM, we need to do dynamic imports to use it from the extension's commonjs code:

```ts
const { Language } = await import("@nomicfoundation/slang")
```

This will require reworking the structure of the supporting classes in Semantic Highlighting and Outline.

antico5:
- Upgraded slang to 1.0.0, which doesnt have native bindings nor optional packages
- Replaced all static imports for dynamic imports. This means many sync functions are now  async.
- Updated all calls to slang with the new apis
- Updated documentSymbols and semanticTokens features
- Installed prettier 3 with an alias (prettier3), to be used on our codebase. Wrote a small script to access it binary. This is necessary to support TS 5.3+ syntax (import attributes in our case).

Closes #632

### DIFF
--- a/server/src/services/initialization/onInitialize.ts
+++ b/server/src/services/initialization/onInitialize.ts
@@ -6,7 +6,6 @@ import {
 } from "vscode-languageserver/node";
 import { ServerState } from "../../types";
 import { tokensTypes } from "../semanticHighlight/tokenTypes";
-import { isSlangSupported } from "../../parser/slangHelpers";
 import { indexWorkspaceFolders } from "./indexWorkspaceFolders";
 import { updateAvailableSolcVersions } from "./updateAvailableSolcVersions";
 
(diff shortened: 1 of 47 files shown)


server/src/services/initialization/onInitialize.ts
```diff
@@ -6,7 +6,6 @@ import {
 } from "vscode-languageserver/node";
 import { ServerState } from "../../types";
 import { tokensTypes } from "../semanticHighlight/tokenTypes";
-import { isSlangSupported } from "../../parser/slangHelpers";
```
kanej, line 9: Is there no equivalent in Slang now?

antico5: I removed it because as I understand, the wasm build works on every os and arch

kanej (approved):
I have reviewed the files and manually tested on:
- mac
- windows
- docker

Both outline and semantic highlighting are working.

LGTM!

PR #641, opened 2025-02-24
Fix build

antico5:
This includes some changes to ensure that the build works. After the revert on 0.8.10, the development branch didn't build correctly for me locally. 

Also refreshed the lockfile with npm 10.9 which uses lockfile version 3


antico5:
Closing this one because the base development branch is buildable - it starts to fail when the lock file is updated to version 3. We can deal with that separately

PR #642, opened 2025-02-24
Rework forge command invocation (broken EPIPE issue)

antico5:
This PR changes the handling of the child process spawned when running `forge`, currently only used for formatting. The broken EPIPE error happens when a process tries to write to the stdin of a dead process.

### DIFF
--- a/server/src/services/formatting/forgeFormat.ts
+++ b/server/src/services/formatting/forgeFormat.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import path from "path";
 import { TextDocument } from "vscode-languageserver-textdocument";
 import { TextEdit } from "vscode-languageserver-types";
@@ -5,6 +6,9 @@ import { URI } from "vscode-uri";
 import { resolveForgeCommand } from "../../frameworks/Foundry/resolveForgeCommand";
 import { Logger } from "../../utils/Logger";
 import { execWithInput } from "../../utils/operatingSystem";
+import { TimeoutError } from "../../utils/errors";
+
+const FORMAT_TIMEOUT = 2000;
 
 export async function forgeFormat(
   text: string,
@@ -19,22 +23,31 @@ export async function forgeFormat(
   logger.trace(`CWD: ${cwd}`);
 
   // Wait for the formatted output
-  const execPromise = await execWithInput(
-    `${forgeCommand} fmt --raw -`,
-    text,
-    { cwd },
-    2000
-  );
-  const formattedText = execPromise.stdout;
+  try {
+    const { stdout: formattedText } = await execWithInput(
+      `${forgeCommand} fmt --raw -`,
+      text,
+      {
+        cwd,
+        timeout: FORMAT_TIMEOUT,
+      }
+    );
 
-  // Build and return a text edit with the entire file content
-  const textEdit: TextEdit = {
-    range: {
-      start: { line: 0, character: 0 },
-      end: document.positionAt(text.length),
-    },
-    newText: formattedText,
-  };
+    // Build and return a text edit with the entire file content
+    const textEdit: TextEdit = {
+      range: {
+        start: { line: 0, character: 0 },
+        end: document.positionAt(text.length),
+      },
+      newText: formattedText,
+    };
 
-  return [textEdit];
+    return [textEdit];
+  } catch (error: any) {
+    if (error.killed) {
+      throw new TimeoutError(FORMAT_TIMEOUT);
+    } else {
+      throw error;
+    }
+  }
 }
--- a/server/src/services/formatting/forgeFormat.ts
+++ b/server/src/services/formatting/forgeFormat.ts
@@ -3,6 +3,7 @@ import path from "path";
 import { TextDocument } from "vscode-languageserver-textdocument";
 import { TextEdit } from "vscode-languageserver-types";
 import { URI } from "vscode-uri";
+import type { ExecException } from "node:child_process";
 import { resolveForgeCommand } from "../../frameworks/Foundry/resolveForgeCommand";
 import { Logger } from "../../utils/Logger";
 import { execWithInput } from "../../utils/operatingSystem";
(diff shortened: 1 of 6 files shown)


server/src/utils/operatingSystem.ts
```diff
@@ -17,6 +17,61 @@ export async function runCmd(cmd: string, cwd?: string): Promise<string> {
...
+        if (writeErr) {
+          reject(`Error writing to stdin: ${writeErr}}`);
+          // Even on error, end the stream to avoid hanging.
+          child.stdin?.end();
+        } else {
+          child.stdin?.end();
+        }
+      });
+    } else {
+      reject("Child process has no writable stdin stream");
```
kanej, line 70: I think we reject with an `Error` in each of these.

server/src/utils/operatingSystem.ts
```diff
@@ -17,6 +17,61 @@ export async function runCmd(cmd: string, cwd?: string): Promise<string> {
...
+    if (timeout !== undefined) {
+      setTimeout(() => {
+        reject("Timed out");
+      }, timeout);
+    }
+
+    // Start the child process
+    const child = exec(command, options, (error, stdout, stderr) => {
+      if (error) {
+        reject(error);
```
kanej, line 47: I think we need to change this to return immediately. I am not sure how node will handle rejecting then continuing on to resolve.

kanej (changes requested):
I think we need to clear the timeout on process exit.
I am tempted to say we should rework the timeout Error, so that when it is caught by the try/catch in `onDocumentFormatting` we report a more specific sentry status for timeout: `{ status: "internal_error", result: null }`. This would let us separate out timeouts from anything else.

kanej:
As a sidepoint - the codecov here is really painful to review. If it is not pulling its weight we should consider turning it off.
What do you think would be required to bring `execWithInput` under test?

antico5:
@kanej addressed all comments. About codecov, I'd be happy to remove it. We can tell on PR reviews what needs and what doesn't need coverage. For execWithInput, I think it's not viable to unit test. It would involve a lot of stubbing and mocking, and the added value would be almost nonexistent. For a non-stub testing case, we'd have to see what's possible in terms of process manipulation from our test case, and it can be tricky to have it be multi platform.

server/src/services/formatting/forgeFormat.ts
```diff
@@ -19,34 +22,32 @@ export async function forgeFormat(
...
+    };
+
+    return [textEdit];
+  } catch (error: any) {
+    if (error.killed) {
+      throw new TimeoutError(FORMAT_TIMEOUT);
+    } else {
+      throw error;
+    }
+  }
```
kanej, line 52:
```suggestion
  } catch (error: unknown) {
    if (error instanceof Error && "killed" in error && error.killed) {
      throw new TimeoutError(FORMAT_TIMEOUT);
    } else {
      throw error;
    }
  }
```
We are trying to avoid `any`, I suspect we will want to add helpers to utils to better support error detection patterns.

server/src/services/formatting/onDocumentFormatting.ts
```diff
@@ -46,11 +47,17 @@ export function onDocumentFormatting(serverState: ServerState) {
...
           );
 
-          return { status: "internal_error", result: null };
+          if (error instanceof TimeoutError) {
+            return { status: "deadline_exceeded", result: null };
+          } else if (error instanceof ForgeResolveError) {
+            return { status: "failed_precondition", result: null };
+          } else {
+            return { status: "internal_error", result: null };
+          }
```
kanej, line 60:
```suggestion
          } else {
            serverState.logger.error(error);
            return { status: "internal_error", result: null };
          }
```

We effectively short-circuit the error handling of `trackTiming`; it would Sentry log the error. I don't think we want to do that for timeouts or failed preconditions - but what do we think of reporting internal errors?
I don't have a strong feeling either way.

server/src/utils/operatingSystem.ts
```diff
@@ -10,13 +10,66 @@ export async function runCmd(cmd: string, cwd?: string): Promise<string> {
...
+      });
+
+      child.once("spawn", () => {
+        stdin.write(input, (error) => {
+          if (error) {
+            reject(error);
+          }
+          stdin.end();
+        });
+      });
```
kanej, line 66:
```suggestion
      child.once("spawn", () => {
        if (!stdin.writable || child.killed) {
          return reject(new Error("Failed to write to unwritable stdin"));
        }

        stdin.write(input, (error) => {
          if (error) {
            reject(error);
          }
          stdin.end();
        });
      });
```

Do we still need a writable/killed guard, or do we think this is covered by the initial spawn step?

server/src/services/formatting/onDocumentFormatting.ts
antico5: As of now, saving a solidity file with syntax errors would cause forge fmt to fail, and that would result in an internal_error. Should I handle the additional case where the program executed but exit code was non zero? and avoid reporting it to sentry ?

server/src/utils/operatingSystem.ts
antico5: I had that guard in place but removed it because I didn't find any way to trigger it. But I guess we can just leave it to be extra sure.

server/src/services/formatting/forgeFormat.ts
antico5: Good point. I just stumbled upon the "ensureError" logic on hardhat. We could use something like that in the future

server/src/services/formatting/onDocumentFormatting.ts
kanej: Ah. No lets leave this reporting out completely then. Forge fmt failing on syntax errors is definitely not worth reporting.

server/src/utils/operatingSystem.ts
kanej: I would have no problem removing it if we know of no way it could be hit.

kanej (approved):

PR #653, opened 2025-03-18
Prepare for slang 0.19

antico5:
Picked up the commits that were reverted previously, and resolved some conflicts.
- Scoped e2e tests project
- Changed modules and target
- Updated typescript and many other dependencies
- Bump minimum vscode version to 1.90, which uses electron 29, which is the first to have node 20+


server/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es2022",
-    "lib": ["es2020"],
```
kanej, line 4: This was inconsistent with the other tsconfig files and the reference config we used for Node 20:

https://www.npmjs.com/package/@tsconfig/node20.

coc/package.json
```diff
@@ -14,6 +14,7 @@
     "ethereum",
     "hardhat"
   ],
+  "type": "commonjs",
```
kanej, line 17: This is to hint to esbuild that we are explicitly commonjs code.

kanej (approved):
I added a few commits:

* bring the server tsconfig inline with other tsconfig files
* target node 20 with esbuild
* be explicit about commonjs and cjs format
* set a minimum node version in our two released packages

@antico5 take a look at the new commits and see what you think.

PR #654, opened 2025-03-21
Upgrade to Sentry 9.6.1

antico5:
- Removed deprecated packages: `sentry/tracing` and `sentry/types`
- Updated sentry/node to latest version, 9.6.1
- Added sentry/core, which is a transient dependency of sentry/node since some functions and types are not re-exported by sentry/node
- Some concepts ceased to exist, for instance Transactions. So I migrated to the new approach which is to use only spans (top level spans are what transactions used to be)
- Some methods e.g. `setTag` now should be invoked directly from Sentry, so our application is a bit more coupled to sentry instead of the abstract Telemetry class. I think it's unnecesary to have such abstraction, and would get rid of it. If we wanted to completely decouple from Sentry it'd be too much overhead for little benefit

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -56,7 +56,6 @@
     "postpackage": "rimraf ./tmp"
   },
   "devDependencies": {
-    "@sentry/types": "6.19.1",
     "@types/prettier": "2.6.0",
     "@types/vscode": "^1.86",
     "eslint": "^7.23.0",
@@ -65,7 +64,7 @@
   },
   "dependencies": {
     "@nomicfoundation/solidity-language-server": "0.6.16",
-    "@sentry/node": "6.19.1",
+    "@sentry/node": "9.6.1",
     "vscode-languageclient": "^7.0.0"
   },
   "contributes": {
--- a/server/src/services/completion/onCompletion.ts
+++ b/server/src/services/completion/onCompletion.ts
@@ -26,6 +26,7 @@ import {
 import { applyEditToDocumentAnalyzer } from "@utils/applyEditToDocumentAnalyzer";
 import { ServerState } from "../../types";
 import { addFrameworkTag } from "../../telemetry/tags";
+import { INTERNAL_ERROR, OK } from "../../telemetry/TelemetryStatus";
 import { ProjectContext } from "./types";
 import { getImportPathCompletion } from "./getImportPathCompletion";
 import { globalVariables, defaultCompletion } from "./defaultCompletion";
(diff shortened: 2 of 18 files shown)


antico5:
@kanej ready for review

client/package.json
```diff
@@ -57,15 +56,14 @@
...
   "devDependencies": {
-    "@sentry/types": "6.19.1",
     "@types/prettier": "2.6.0",
     "@types/vscode": "^1.90",
     "rimraf": "3.0.2",
     "prettier": "2.5.1"
   },
   "dependencies": {
     "@nomicfoundation/solidity-language-server": "0.8.20",
-    "@sentry/node": "6.19.1",
```
kanej, line 68: Should the client change to be `@sentry/browser`?
This is based on https://docs.sentry.io/platforms/javascript/best-practices/shared-environments/

client/src/extension.ts
```diff
@@ -18,33 +18,38 @@ const SENTRY_DSN =
...
 
-  return {
-    isReady: () => !!extensionState?.client?.initializeResult,
-  };
+    return {
+      isReady: () => !!extensionState?.client?.initializeResult,
+    };
+  } catch (error) {
+    extensionState.logger.error(error);
+    throw error;
```
kanej, line 51: Is this a direct replacement of the global handler on the client-side? Or are there additional changes needed to replicate the global handler?

client/src/telemetry/SentryClientTelemetry.ts
```diff
@@ -18,33 +20,57 @@ export class SentryClientTelemetry implements Telemetry {
...
+      (defaultIntegration) => {
+        return !["BrowserApiErrors", "Breadcrumbs", "GlobalHandlers"].includes(
+          defaultIntegration.name
+        );
+      }
+    );
+
+    const client = new Sentry.NodeClient({
       dsn: this.dsn,
-      tracesSampleRate: this.extensionState.env === "development" ? 1 : 0.001,
```
kanej, line 23: We don't need rate sampling because we don't use spans (or other sampling) techniques on the client-side?

server/src/services/completion/onCompletion.ts
```diff
@@ -38,44 +39,41 @@ export const onCompletion = (serverState: ServerState) => {
...
+          (doc) => resolveChangedDocText(params, doc)
+        );
+
+      if (!found || !documentAnalyzer || !document) {
+        logger.trace(
+          `Error editing and analyzing doc within onCompletion: ${errorMessage}`
         );
 
-        return { status: "ok", result: completions };
+        return { status: INTERNAL_ERROR, result: null };
```
kanej, line 55: Why not keep it as a `FAILED_PRECONDITION`?

client/src/extension.ts
antico5: In this particular case, it is only for the initialization code.  I've wrapped all entry points where the extension could fail on the client side and capture the exception on sentry. We can't use the global handler because of the shared environment, so we have to be explicit on every point we want to capture an exception

client/package.json
antico5: Since our extension runs on a node js environment, I'm confident the @sentry/node package is the more apropiate one. Checking the documentation page, it may be confusing because sometimes it refers to browser extensions themselves, but vscode extensions are treated as a separate concept. Although if there is any potential advantage on the `@sentry/browser` package that we may be interested in, I can dive deeper into it

client/src/telemetry/SentryClientTelemetry.ts
antico5: That's correct

server/src/services/completion/onCompletion.ts
antico5: Good catch,  this looks like a mistake

client/package.json
kanej: Ah okay, so I have misunderstood.
It is not that we should swap to the reduced browser client, but that we should use the initialization techniques (which they demonstrate with the browser client).

kanej: I am happy to go with the node version and determine from Sentry whether it achieves our goals. Specifically:

* No uncaught exceptions from other extensions
* ideally no phantom extensions in our sentry reports.

antico5: When working on this I've created a dummy extension alongside that periodically crashed to ensure errors from that extension wouldn't get in the way. For process unhandled errors, there's a special check that would report them only if our extension is on the stack trace (line 58 on SentryClientTelemetry)

server/src/services/initialization/indexWorkspaceFolders.ts
```diff
@@ -77,34 +76,32 @@ export async function indexWorkspaceFolders(
...
-    solFileUris = await scanForSolFiles(
-      logger,
-      workspaceFileRetriever,
-      topLevelWorkspaceFolders
-    );
-
-    // Index sol files, and associate the matching project
-    await indexSolidityFiles(serverState, solFileUris);
-    span?.finish();
+    await startSpan({ name: "findSolidityFiles" }, async () => {
```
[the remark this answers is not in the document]
kanej: On looking deeper this is `logger.trackTime` rather than `telemetry.trackTiming`. I am unsure about the value of `logger.trackTime` but it shouldn't be calling out to Sentry.

server/src/telemetry/types.ts
```diff
@@ -18,15 +17,14 @@ export interface Telemetry {
   captureException(err: unknown): void;
   trackTiming<T>(
     taskName: string,
-    action: (transaction: Transaction) => Promise<TrackingResult<T>>,
+    action: (span: Span) => Promise<TrackingResult<T>>,
```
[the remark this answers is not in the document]
antico5: Done, removed from `trackTiming`, `trackTimingSync`, and the other smaller changes

client/src/telemetry/SentryClientTelemetry.ts
```diff
@@ -18,33 +20,57 @@ export class SentryClientTelemetry implements Telemetry {
...
-          );
-        }),
+      release: `${this.extensionState.name}@${this.extensionState.version}`,
       beforeSend: (event) => (isTelemetryEnabled() ? event : null),
     });
+    Sentry.getGlobalScope().setUser({ id: this.extensionState.machineId });
+    Sentry.getGlobalScope().setTag("component", "ext");
+    Sentry.getGlobalScope().setTag("isHandled", true);
+
+    client.init();
```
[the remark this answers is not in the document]
kanej: Lets leave it with using the global scope for now - and monitor how it performs in production. If we start seeing the return of external exceptions we will revisit using a custom scope.

PR #656, opened 2025-03-25
Improve parent child process

### LINKED ISSUE #655: Improve interprocess communication with Hardht worker process
To help guard against potential EPIPE issues when dealing with validation in Hardhat processes there are several moves we want to make.

antico5:
- Added error handling to all calls to `send`, so any `ERR_IPC_CHANNEL_CLOSED` errors will be handled
- Exit on child process once parent process exits.
- Wrap with try/catch on some project methods that may throw errors

Closes #655

### DIFF
--- a/server/src/frameworks/Hardhat/HardhatProject.ts
+++ b/server/src/frameworks/Hardhat/HardhatProject.ts
@@ -216,7 +216,7 @@ export class HardhatProject extends Project {
 
     return Promise.race([
       workerPromise,
-      this._requestTimeout("fileBelongs"),
+      this._requestTimeout("resolveImportPath"),
     ]) as Promise<string>;
   }
 
@@ -252,7 +252,13 @@ export class HardhatProject extends Project {
       if (error) {
         this.workerStatus = WorkerStatus.ERRORED;
         this.logger.error(`Error sending message to hardhat worker: ${error}`);
-        throw error;
+
+        if ("requestId" in message) {
+          const messageWithId: { requestId: number } = message as any;
+          this._callErrorForRequest(messageWithId.requestId, error);
+        } else {
+          throw error;
+        }
       }
     });
   }
@@ -351,17 +357,7 @@ export class HardhatProject extends Project {
   }
 
   private _handleErrorResponse(msg: ErrorResponseMessage) {
-    const errorFunction = this._onError[msg.requestId];
-
-    if (errorFunction === undefined) {
-      this.logger.error(
-        `Error function not found for request id ${msg.requestId}`
-      );
-    } else {
-      delete this._onError[msg.requestId];
-      delete this._onResponse[msg.requestId];
-      errorFunction(msg.error);
-    }
+    this._callErrorForRequest(msg.requestId, msg.error);
   }
 
   private _handleFileBelongsResponse(msg: FileBelongsResponse) {
@@ -432,6 +428,18 @@ export class HardhatProject extends Project {
       };
 
       throw error;
+    }
+  }
+
+  private _callErrorForRequest(requestId: number, error: any) {
+    const errorFunction = this._onError[requestId];
+
+    if (errorFunction === undefined) {
+      this.logger.error(`Error function not found for request id ${requestId}`);
+    } else {
+      delete this._onError[requestId];
+      delete this._onResponse[requestId];
+      errorFunction(error);
     }
   }
 }
--- a/server/src/frameworks/Hardhat/HardhatProject.ts
+++ b/server/src/frameworks/Hardhat/HardhatProject.ts
@@ -114,6 +114,8 @@ export class HardhatProject extends Project {
 
       this.workerProcess.on("error", (err) => {
         this.logger.info(`Worker exited with error: ${err}`);
+        this.workerLoadFailureReason = `Child process couldnt be spawned for ${this.id()}`;
+        this.workerStatus = WorkerStatus.ERRORED;
       });
 
       this.workerProcess.on("exit", async (code) => {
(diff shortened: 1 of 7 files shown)


.github/codecov.yml
```diff
@@ -1,12 +1,4 @@
...
-        threshold: 50%
-        base: development
-    patch:
-      default:
-        threshold: 50%
-    changes:
-      default:
-        threshold: 10%
+    project: off
+    patch: off
```
kanej, line 4: We can revisit how we approach codecoverage in the project later. I don't like having to ignore red fails in the CI - better to admit reality and turn off the check.

server/src/frameworks/Hardhat/HardhatProject.ts
```diff
@@ -253,6 +255,17 @@ export class HardhatProject extends Project {
     return resolveActionsFor(this.serverState, diagnostic, document, uri);
   }
 
+  private _sendToChild(message: Message) {
+    this._assertWorkerExists();
+    this.workerProcess!.send(message, (error) => {
```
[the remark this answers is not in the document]
antico5: Awesome, I didn't know about such assertion functions!

server/src/frameworks/Hardhat/HardhatProject.ts
```diff
@@ -253,6 +243,32 @@ export class HardhatProject extends Project {
     return resolveActionsFor(this.serverState, diagnostic, document, uri);
   }
 
+  private _sendToChild(message: Message) {
+    this._assertWorkerExists();
+    this._assertWorkerIsRunning();
```
kanej, line 248: We only want to be sending if the child process is considered running; if it is initializing, errored or stopped we should not be sending a message.

server/src/frameworks/Hardhat/HardhatProject.ts
```diff
@@ -253,6 +243,32 @@ export class HardhatProject extends Project {
...
+    this._assertWorkerConnected();
```
kanej, line 249: I added an assertion to check that the child process is still connected.

server/src/frameworks/Hardhat/HardhatProject.ts
```diff
@@ -253,6 +255,17 @@ export class HardhatProject extends Project {
     return resolveActionsFor(this.serverState, diagnostic, document, uri);
   }
 
+  private _sendToChild(message: Message) {
+    this._assertWorkerExists();
+    this.workerProcess!.send(message, (error) => {
+      if (error) {
+        this.workerStatus = WorkerStatus.ERRORED;
+        this.logger.error(`Error sending message to hardhat worker: ${error}`);
```
kanej, line 254: My reading of the docs around adding an error handler to `send`, means that instead of an `error` event being triggered now just this callback will happen.
From the doc string of send:
> If no callback function is provided and the message cannot be sent, an'error' event will be emitted by the ChildProcess object. This can happen, for instance, when the child process has already exited.

The only difference based on this logic is:
1. we don't set the `workerLoadFailureReason`, which makes sense we have erroneously been assuming that any error raised was one during initial boot of the process
2. We send a sentry error (rather than just info) every time this happens now

I was tempted to say we switch this to only `this.logger.info` matching the current behaviour. But this code path should be very rare and is valuable to our understanding, so lets leave like this.

server/src/frameworks/Hardhat/HardhatProject.ts
kanej: I removed the `_assertWorkerIsRunning` check. Looking through the code I can't predict if there are places where we don't make this assumption.

server/src/frameworks/Hardhat/HardhatProject.ts
kanej: We always want to check whether the child process is connected before calling send.

server/src/frameworks/Hardhat/HardhatProject.ts
```diff
@@ -253,6 +247,37 @@ export class HardhatProject extends Project {
...
+    this._assertWorkerConnected();
+
+    this.workerProcess.send(message, (error) => {
+      if (error) {
+        this.workerStatus = WorkerStatus.ERRORED;
+        this.logger.error(`Error sending message to hardhat worker: ${error}`);
+
+        if ("requestId" in message) {
+          const messageWithId: { requestId: number } = message as any;
+          this._callErrorForRequest(messageWithId.requestId, error);
```
kanej, line 261: I enhanced this to replicate the cleanup logic we use for an error message against a specific request id.
If there is no `requestId` we fall back as before.

server/src/frameworks/shared/crawlDependencies.ts
```diff
@@ -55,13 +55,19 @@ export async function getDependenciesAndPragmas(
...
+        importName
+      );
 
-    if (resolvedPath !== undefined) {
-      resolvedImports.push({ importName, resolvedPath });
+      if (resolvedPath !== undefined) {
+        resolvedImports.push({ importName, resolvedPath });
+      }
+    } catch (error) {
+      project.serverState.logger.error(
```
kanej, line 68: Why wouldn't we fail the validation if this fails? Why swallow the error (though reporting it)?

kanej (changes requested):
I think we are there on this PR with the exception of my question around catching the error in `crawlDependencies`.

server/src/frameworks/shared/crawlDependencies.ts
kanej: Out of band we decided to remove this try/catch.

server/src/services/rename/onRename.ts
```diff
@@ -36,7 +36,13 @@ function rename(
   position: VSCodePosition,
   newName: string
 ): WorkspaceEdit {
-  solFileEntry.project.invalidateBuildCache();
+  try {
+    solFileEntry.project.invalidateBuildCache();
```
kanej, line 40: We are going to revert this try/catch.

PR #665, opened 2025-04-02
bump node language server and client

kanej:
To deal with EPIPE errors we want to update to the latest versions of the vscode language server libraries that underlie our Language Server.
We have previously updated the language server side, this PR updates the client as well.

The client saw breaking changes around v8: https://github.com/Microsoft/vscode-languageserver-node?tab=readme-ov-file#3170-protocol-800-json-rpc-800-client-and-800-server.

In this PR we:

1. Update the library versions
2. Swap in the client code to using `.start()` instead of `.onReady()` - this needs to be analyzed to ensure we are not changing initialization order with multiple starts
3. Update the tests to match the new async apis

## Manual Testing

I have done a quick test round with the smoke test repo and a bundled version of these changes.

## Reviewer Suggestions

Take this one commit at a time. The main changes are commit 2 which swaps `.start()` for `.onReady()` and wires through the relevant awaits (or voids if that is appropriate).

A read through the changelog for the client is necessary as a check: https://github.com/microsoft/vscode-languageserver-node?tab=readme-ov-file#3170-protocol-800-json-rpc-800-client-and-800-server

I have not changed deactive in `extension.ts` but believe we are returning an appropriate promise that meets this requirement:

> Extensions should now implement a deactivate function in their extension main file and correctly return the stop promise from the deactivate call

### DIFF
--- a/client/src/setup/setupLanguageServerHooks.ts
+++ b/client/src/setup/setupLanguageServerHooks.ts
@@ -10,11 +10,13 @@ import { isTelemetryEnabled } from "../utils/telemetry";
 import { setupIndexingHooks } from "./setupIndexingHooks";
 import { setupValidationJobHooks } from "./setupValidationJobHooks";
 
-export function setupLanguageServerHooks(extensionState: ExtensionState) {
-  startLanguageServer(extensionState);
+export async function setupLanguageServerHooks(extensionState: ExtensionState) {
+  return startLanguageServer(extensionState);
 }
 
-const startLanguageServer = (extensionState: ExtensionState): void => {
+const startLanguageServer = async (
+  extensionState: ExtensionState
+): Promise<void> => {
   const { logger } = extensionState;
 
   // The debug options for the server.
@@ -78,7 +80,7 @@ const startLanguageServer = (extensionState: ExtensionState): void => {
 
   setupIndexingHooks(extensionState, client);
 
-  setupValidationJobHooks(extensionState, client)
+  await setupValidationJobHooks(extensionState, client)
     .then((disposable) => extensionState.listenerDisposables.push(disposable))
     .catch((err) => extensionState.logger.error(err));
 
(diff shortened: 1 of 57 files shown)


client/src/setup/setupLanguageServerHooks.ts
```diff
@@ -78,15 +80,16 @@ const startLanguageServer = (extensionState: ExtensionState): void => {
...
     .then((disposable) => extensionState.listenerDisposables.push(disposable))
     .catch((err) => extensionState.logger.error(err));
 
   const notifyTelemetryChanged = () => {
     const telemetryEnabled = isTelemetryEnabled();
 
     extensionState.telemetryEnabled = telemetryEnabled;
-    client.sendNotification("custom/didChangeTelemetryEnabled", {
+
+    void client.sendNotification("custom/didChangeTelemetryEnabled", {
```
alcuadrado, line 92: As a follow up, I think we should have a helper that does this `void` trick, which should have docs about why it's done.

alcuadrado:
I think this looks good. I have just three comments:

- The pinned versions are pretty hard to analyze. It took me some time to be sure that they are aligned between each other. I don't think there's much we can do, as they are pinning upstream. Maybe a github workflow check?
- The `void` for floating promises is correct in this case, as they suggest, but I think we'll completely forget about it. I'd rather move it to a helper.
- The `deactivate` promise doesn't really return the promise of `stop`. It's return type is different, because it's a `Promise.all()`. I think we can either do `return Promise.all([clientStopPromise, telemetryClosePromise]).then(([ret]) => ret);`, or run one after the other.

kanej:
> I think this looks good. I have just three comments:
> 
>     * The pinned versions are pretty hard to analyze. It took me some time to be sure that they are aligned between each other. I don't think there's much we can do, as they are pinning upstream. Maybe a github workflow check?
> 
>     * The `void` for floating promises is correct in this case, as they suggest, but I think we'll completely forget about it. I'd rather move it to a helper.
> 
>     * The `deactivate` promise doesn't really return the promise of `stop`. It's return type is different, because it's a `Promise.all()`. I think we can either do `return Promise.all([clientStopPromise, telemetryClosePromise]).then(([ret]) => ret);`, or run one after the other.

Is stop returns a `Promise<void>`, is the return type of `Promise<[void, boolean]>` not equivalent?

alcuadrado:
Oh, I think you are right. It wouldn't compile otherwise.

antico5 (approved):

PR #667, opened 2025-04-03
chore: add breadcrumbs to sentry events on uncaught error events

antico5:
Patch `EventEmitter` `init` and `emit` functions to pick up stack traces and append them as breadcrumbs on process' uncaught exceptions. This is an effort to have more context on unhandled EPIPE errors


kanej (approved):
I don't have anything to add.
My assumption is, this won't cause a memory leak because we do not hold onto thrown errors - so the weakmaps are being continuously cleaned up.
We should do a manual run through and check that the background processes are the same as before this change. So run through the manual test suite locally, with `Developer Process Explorer` open, then apply this change and run through again.

PR #679, opened 2025-05-08
Hardhat 3 projects support

antico5:
This PR is temporarily against the slang upgrade branch because it's based on top of that.

**Included in the PR**:
  - Identifying, indexing hardhat 3 projects
  - Validation on hardhat 3 projects
  - Analysis (all language features) on hh3 projects

**Missing from the PR**:
  - Import completions of npm packages and remapped paths
  - Hardhat.config reloading
  - Validation on npm contracts
  - using hh3 for compilation

### General
  - No IPC needed like in hardhat2 projects, since multiple HRE can be spawned on the same process without issue
  - Added hardhat3 as a dependency to the server so we can use the types. It doesn't add up on the bundle size since it's stripped out by esbuild
  - The "Hardhat" project type was renamed to Hardhat 2, but mostly internally, like on file names, class names etc
  - We are using `importUserConfig`, `createHardhatRuntimeEnvironment`, and `DependencyResolver` from the project's hardhat package.
  - We insert an ad-hoc plugin that defines the `readSourceFile` solidity hook, so it can read from open documents before falling back to disk files.
  - Ported over the AsyncMutex class from Hardhat. Maybe we can move it to hardhat-utils eventually ?
  - Added the `read-pkg` package as dependency to read package.json files to determine hardhat versions
  - I introduced some helper functions to ensure windows absolute paths are normalized, i.e. uppercase drive letters.

### Indexing
  - Project detection is based on the version range of the hardhat package in package.json
  - I didn't add an extra indexer, I just changed the existing one to return both hardhat 2 and hardhat 3 project instances
  - Hardhat3Project has the highest priority, so e.g. it will take precedence even if a project folder has a foundry.toml file
  - During initialization, an empty call is made to `getCompilationJobs` so all compilers are downloaded. This is to avoid a single validate() call to take potentially > 10s
  - All solidity files under the project root are considered to belong to the hardhat3 project. Those under the `path.sources` directory are marked as local to be eagerly analyzed. Currently nested projects are not supported, so we have to think if it's a possible scenario.
  - If a hardhat.config file is found, but any unexpected errors happen while trying to determine if it's hh3, it defaults to hh2, so there should be no regressions on that side.
  - For a project to be correctly initialized, npm modules need to be installed. Otherwise no validation/multi-file analysis can take place

### Validation
  - I'm mostly passing through to build system getCompilationJobs.
  - If there's an unresolved import error (from hardhat, not solc), like we do for hh2 projects we generate a diagnostic for the import line as well as a status item linking to the file that has the import error, since that prevents any kind of compilation.

### Analysis
  - An imports cache is in place to avoid updating the dependency graph if a file's imports didn't change between edits
  - We create and maintain a custom dependency graph (LSPDependencyGraph) that is updated when reading, creating or removing files. This is used to quickly resolve imports, since once it's built it's just a map lookup. We use the DependencyResolver from hardhat to build it.
  - I introduced a new 'hook' on the analysis process, in which project instances get called before the analysis runs. It's called `Project.preAnalyze`. This was needed because there was no callback from the analyzer to the project besides the resolveImport, and this was needed to preparate for the subsequent calls (i.e. updating the dependency graph)

### LSPDependencyGraph
  - A custom implementation that has several ResolvedFiles, indexed by absolute path.
  - Each node has their dependencies and dependants
  - Unresolved dependencies are tracked for efficiency purposes
  - DependencyResolver instance is recreated many times because the implementation has caching on a file basis. So if a file changes the content wouldn't be updated if using always the same instance

### Testing
  - All protocol tests have been copied and adapted from hh2 to hh3, and new ones added
  - Added tests on all frameworks for the custom/file-indexed notification, which shows on the UI the project associated with a given solidity file
  - Added a test case for hh3 when packages are not installed (error status item)
  - Added a test suite for all types of import resolution on hardhat 3:
    - from local file
      - to local file through relative
      - to local file direct
      - to local file through remapping
      - to npm package direct
      - to npm package through remapping
      - to npm package with exports direct
      - to npm package with exports through remapping
    - from npm package
      - to same pkg relative
      - to same pkg direct
      - remappings should not work
      - to other npm package direct
      - to other npm package direct with exports


antico5:
remove read-pkg package

server/src/frameworks/Hardhat/Hardhat3/Hardhat3Project.ts
```diff
@@ -0,0 +1,489 @@
...
+import { LSPDependencyGraph } from "./LSPDependencyGraph";
+
+export class Hardhat3Project extends Project {
+  public priority = 5;
+
+  protected logger: Logger;
+  protected initializeError?: string; // Any error message that happened during initialization
+  protected hre?: HardhatRuntimeEnvironment; // HRE instance of this project
+  protected dependencyGraph?: LSPDependencyGraph; // Dependency graph built/maintained for import resolution on analysis
+  protected openDocuments: OpenDocuments = []; // Stored on buildCompilation calls, needed for the readSourceFile hook override
```
antico5, line 48: get rid of this

test/protocol/test/textDocument/definition/hardhat3/imports.test.ts
```diff
@@ -0,0 +1,178 @@
...
+
+      const location = await client.findDefinition(toUri(localFilePath), makePosition(7, 2))
+
+      expect(location).to.deep.equal({
+        uri: toUri(getTestPackagePath('pkg_with_exports_1', 'src/C.sol')),
+        range: makeRange(6, 9, 6, 10),
+      })
+    })
+
+    test('to npm package (with exports) through remappings', async () => {
```
kanej, line 116: Nice. These are clear and will make good regression tests when the HH3 build system changes come through.

server/src/frameworks/Hardhat/Hardhat3/LSPDependencyGraph.ts
```diff
@@ -0,0 +1,256 @@
...
+            );
+
+            if (importedFile.fsPath === newFileAbsPath) {
+              this.unresolvedImports.get(unresolvedAbsPath)?.delete(importPath);
+
+              this.#addDependency(potentialDependant, importedFile, importPath);
+              this.#addDependant(potentialDependant, importedFile, importPath);
+            }
+          } catch (error) {
+            continue;
```
kanej, line 151: Can you talk me through this error condition?

kanej: We talked through this out of band.

PR #683, opened 2025-06-06
Sentry event anonymization

### LINKED ISSUE #485: Improve exception handling user data scrubbing
Sentry data collection uses the anonymized machineId when reporting exceptions. However, because we capture unexpected exceptions these can capture user home directory names, if there is a filesystem exception for instance.

The first step is a further investigation of the options. We should look at how we are dealing with anonymization in Hardhat 3 and determine what should be ported/expanded upon.

That investigation should lead to a design doc.

antico5:
Closes #485 

All potential user-identifying information that may be present in error events or performance traces are now scrubbed before being sent to Sentry.

File paths and URLs are replaced with `<user_file>` strings. For the extension's internal files, we scrub up to the the extension root path, and keep the internal file subpath for debugging purposes. This will result in paths looking like `<extension_root>/server/out/index.js`. This detection of internal files additionally is only applied to `.js` files, so we default on the safe side of scrubbing most file paths (which include .sol files).

Additionally some default fields that may be populated by the SDK are cleared as well:
- User email
- User name
- Server name

Some fields may differ from what has been specified in the design document, and this is because the `Event` structure changed significantly from Sentry v6 to Sentry v9. Since this change is subject to the Sentry 9 upgrade, we only care about the new event format.

### DIFF
--- a/server/src/telemetry/anonymization.ts
+++ b/server/src/telemetry/anonymization.ts
@@ -1,0 +1,103 @@
+/* eslint-disable @typescript-eslint/strict-boolean-expressions */
+import { Event, Exception } from "@sentry/core";
+
+const ANONYMIZED_FILE = "<user-file>";
+
+function anonymizeBreadcrumbs(
+  breadcrumbs?: Event["breadcrumbs"]
+): Event["breadcrumbs"] {
+  if (breadcrumbs === undefined) {
+    return undefined;
+  }
+
+  return breadcrumbs.map((breadcrumb) => {
+    return {
+      ...breadcrumb,
+      message: anonymizeString(breadcrumb.message),
+    };
+  });
+}
+
+function anonymizeStackTrace(
+  stacktrace?: Exception["stacktrace"]
+): Exception["stacktrace"] {
+  if (stacktrace?.frames === undefined) {
+    return stacktrace;
+  }
+
+  return {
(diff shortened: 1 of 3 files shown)


server/src/telemetry/anonymization.ts
```diff
@@ -0,0 +1,103 @@
...
+    ...event,
+    message: anonymizeString(event.message), // Scrub message
+    server_name: undefined, // Remove server_name
+    breadcrumbs, // Anonimized breadcrumbs
+    exception, // Anonimized exception
+    user, // Anonimized user
+  };
+
+  return scrubbedEvent;
+}
```
kanej, line 86:
```suggestion
export function anonymizeEvent<T extends Event>(event: T): T {
  const breadcrumbs = anonymizeBreadcrumbs(event.breadcrumbs);
  const exception = anonymizeExceptions(event.exception);
  const user = anonymizeUser(event.user);

  const scrubbedEvent: T = {
    ...event,
    message: anonymizeString(event.message), // Scrub message
    server_name: undefined, // Remove server_name
    breadcrumbs, // Anonimized breadcrumbs
    exception, // Anonimized exception
    user, // Anonimized user
  };

  return scrubbedEvent;
}
```

This change would allow the removal of the `as` casting in ServerSentyTelemetry.

PR #685, opened 2025-06-17
Sentry event anonymization for extension client

antico5:
Duplicated server anonimization logic. We may get rid of this duplication once we set up a monorepo structure


kanej (approved):
LGTM!

PR #687, opened 2025-06-23
Reconfirm telemetry opt-in

antico5:
This PR resets the prompt for telemetry opt-in. A small change was added so if the user closes the dialog without choosing Accept nor Decline, it doesn't record anything and will be prompted again on next startup.


kanej:
We took another approach.

PR #688, opened 2025-06-25
WIP - Upgrade to new dependency resolution

antico5:
This PR should update the hardhat version on both the server and the hardhat3 protocol test project. CI will fail because I was using the custom branch on hardhat locally. All tests are passing on linux.


server/src/frameworks/Hardhat/Hardhat3/LSPDependencyGraph.ts
```diff
@@ -29,7 +29,13 @@ export class LSPDependencyGraph {
 
       const filesToProcess: ResolvedFile[] = [];
 
-      const resolvedFile = await resolver.resolveProjectFile(absPath);
+      const resolveResult = await resolver.resolveProjectFile(absPath);
+
+      if (resolveResult.success !== true) {
+        throw resolveResult.error;
```
antico5, line 35: Use HardhatError + exported helpers

PR #690, opened 2025-07-16
Tag vscode version on google analytics

antico5:
Besides this change, I already setup the custom dimension on our GA4 property

### DIFF
--- a/server/src/services/initialization/onInitialize.ts
+++ b/server/src/services/initialization/onInitialize.ts
@@ -23,7 +23,7 @@ export const onInitialize = (serverState: ServerState) => {
       extensionVersion,
       workspaceFolders,
       clientName,
-      vscodeVersion,
+      clientVersion,
     } = getSessionInfo(params);
 
     updateServerStateFromParams(serverState, params);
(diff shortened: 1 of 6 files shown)


server/src/services/initialization/onInitialize.ts
```diff
@@ -119,14 +123,16 @@ function getSessionInfo(params: InitializeParams) {
 
   const workspaceFolders = params.workspaceFolders || [];
 
-  const clientName = params.clientInfo?.name;
+  const clientName: string | undefined = params.clientInfo?.name;
+  const clientVersion: string | undefined = params.clientInfo?.version;
```
kanej, line 127: In reviewing I noticed that `clientInfo` has a version.
I pushed a commit that swaps out `vscodeVersion` and instead uses the `clientVersion` here.
To test I added both the clientName and clientVersion to the output logging on the lsp.

kanej (approved):
Looks good. I switched to `clientVersion` for the property name and pulled it from the clientInfo. Take a look and see what you think.

PR #694, opened 2025-07-23
Adapt to new getCompilationJobs api

antico5:
Now supports hardhat 3 projects with version >= next.24


kanej (approved):
I ran a manual test with the lastest Alpha and can confirm I get navigation, inline errors and quickfixes.

Lets get this merged.

PR #696, opened 2025-08-05
chore: update Node versions used in CI test matrix

kanej:
The node version used by the extension is now constrained to the version included with VSCode/Cursor. Hence we are testing against those explicit versions.


antico5 (approved):
I added a commit for making dependency resolution tests more deterministic (pnpm was making it messy). Tests are passing now

PR #727, opened 2026-02-10
drop node 20 support

kanej:
Cursor has been updated to VSCode 1.105.1 in the latest version. As such we no longer have to include support for Node 20 in our test matrix.

This PR updates the minimum supported version of VSCode for the extension to `1.101.0` the first version with Node 22 support. It also drops Node 20 from our test matrix.


kanej:
We need to refresh this based on current versions.

PR #731, opened 2026-03-14
fix error with pre-release solc

ssadler:
Pre release solc (can happen if version is not specified for example in foundry.toml) breaks the client because it emits:

```json
    {
      component: 'general',
      errorCode: '3805',
      formattedMessage: 'Warning: This is a pre-release compiler version, please do not use it in production.\n' +
        '\n',
      message: 'This is a pre-release compiler version, please do not use it in production.',
      severity: 'warning',
      type: 'Warning'
    }
```

Causing `Cannot read properties of undefined (reading 'file')` in `OutputConverter.ts`, because it has no `sourceLocation`.


kanej:
Closing in preference of https://github.com/NomicFoundation/hardhat-vscode/pull/737.

PR #736, opened 2026-05-24
fix: add fallback mechanisms for hover and find references

phpmac:
## Summary

The upstream `usedef` module has incomplete `typeNodes` connections, causing **hover** and **find references** to return `null` on many symbols, while **goToDefinition** works correctly via parent chains.

This PR adds multi-layer fallback mechanisms to both features.

### Problem

In the current implementation:
- `goToDefinition` works reliably because it follows the `parent` chain in the analyzer tree
- `hover` depends on `typeNodes[0]` which is frequently `undefined` due to incomplete `usedef` resolution
- `find references` depends on `definitionNode.children` which is often incomplete for the same reason

#### Hover (`onHover.ts`) - 4-layer fallback

1. **`findDefinitionNodeByPosition`** (reuse goToDefinition's parent chain lookup) - covers most cases
2. **`typeNodes[0]`** (original logic) - still works for declaration sites
3. **orphanNodes search** - matches by position in nodes not attached to the main tree
4. **Text-based name search** - extracts identifier from source text at cursor position, then searches the entire analyzerTree by name

#### Find References (`onReferences.ts`) - 2-layer fallback

1. **`findReferencesFor`** (original children-based logic)
2. **`searchReferencesByName`** - recursive tree walk matching `getName()` / `getAliasName()` when children result is empty

### Testing

Tested using the LSP server directly (via Claude Code's LSP integration) with two Solidity test files:

**Counter.sol** (simple contract):
| Test | Position | Result |
|------|----------|--------|
| `uint256 public number` declaration | line 5 | `uint256 public number` |
| `function setNumber` declaration | line 7 | `function setNumber(uint256 newNumber) public` |
| `function increment` declaration | line 11 | `function increment() public` |
| `number` reference in assignment | line 8 | `uint256 public number` |

**TestLSP.sol** (complex contract with struct/enum/event/error/mapping/modifier):
| Test | Position | Result |
|------|----------|--------|
| `uint256 public count` declaration | line 5 | `uint256 public count` |
| `struct UserInfo` declaration | line 8 | `struct UserInfo { string name; uint256 balance; }` |
| `enum Status` declaration | line 13 | `enum Status { Active, Paused, Stopped }` |
| `event CountChanged` declaration | line 15 | `event CountChanged(uint256 newCount)` |
| `function increment` declaration | line 37 | `function increment() public` |
| `count` reference in `count++` | line 38 | `uint256 public count` |

### Limitations

- The text-based name search (fallback layer 4) matches by name only without scope awareness. In practice this rarely triggers because layer 1 (`findDefinitionNodeByPosition`) resolves most cases.
- This is a pragmatic patch for the v1 engine. The v2 rewrite may address the root cause.

## Test plan

- [x] Hover on variable declarations returns type + name
- [x] Hover on function declarations returns full signature
- [x] Hover on struct/enum/event declarations returns definition
- [x] Hover on variable references resolves to the declaration
- [x] goToDefinition still works (unaffected)
- [ ] Find references works with fallback (tested manually, no automated test)

### DIFF
--- a/server/src/services/hover/onHover.ts
+++ b/server/src/services/hover/onHover.ts
@@ -5,7 +5,8 @@ import {
 } from "@analyzer/utils/typeGuards";
 import { getParserPositionFromVSCodePosition } from "@common/utils";
 import { HoverParams, Hover } from "vscode-languageserver/node";
-import { ISolFileEntry, IdentifierNode, MemberAccessNode, Node } from "@common/types";
+import { TextDocument } from "vscode-languageserver-textdocument";
+import { ISolFileEntry, IdentifierNode, MemberAccessNode, Node, Position } from "@common/types";
 import { onCommand } from "@utils/onCommand";
 import { ServerState } from "../../types";
 import { astToText } from "./utils/astToText";
@@ -18,8 +19,8 @@ export function onHover(serverState: ServerState) {
         serverState,
         "onHover",
         params.textDocument.uri,
-        (documentAnalyzer) =>
-          findHoverForNodeAtPosition(documentAnalyzer, params)
+        (documentAnalyzer, document) =>
+          findHoverForNodeAtPosition(documentAnalyzer, document, params)
       );
     } catch (err) {
       serverState.logger.error(err);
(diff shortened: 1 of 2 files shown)


kanej:
I made some modifications in, but the core of this PR was merged here: https://github.com/NomicFoundation/hardhat-vscode/pull/746

PR #743, opened 2026-08-18
switch to pnpm

kanej:
Switch to pnpm as the package manager for the repo, bringing it inline with hardhat.

This PR switches out the package manager and updates the config and CI to match. We retain `npm` as part of the bundling process.

Some small tweaks/fixes were done as no longer having hoisted shared node modules showed up a small number of leaks.

## Review

> [!TIP]
> This really needs to be reviewed a commit at a time!


server/src/utils/operatingSystem.ts
```diff
@@ -40,7 +40,7 @@ export async function execWithInput(
       if (error) {
         reject(error);
       } else {
-        resolve({ stdout, stderr });
+        resolve({ stdout: stdout.toString(), stderr: stderr.toString() });
```
kanej, line 43: This was caused by the update to the @node/types (or more correctly the correct hoisting of the existing types).

client/scripts/bundle.js
```diff
@@ -223,6 +223,11 @@ async function main() {
     })
   );
 
+  // Deliberately npm, not pnpm: this throwaway package outside the workspace
+  // exists to produce a flat `node_modules` holding every platform build of
+  // solidity-analyzer, which `.vscodeignore` then allow-lists into the vsix.
+  // `--force` is what makes npm install optional dependencies for platforms
+  // other than this one, and vsce packages real files, not symlinks.
```
kanej, line 230: Lets look at updating the entire bundling process later, and stick with npm here for now.

alcuadrado (approved):

PR #748, opened 2026-08-20
chore: bump vscode languageserver

kanej:
The language server packages move to their 10.x line: `vscode-languageclient` and `vscode-languageserver` 9.0.1 to 10.1.0 (in the client and in `test/e2e`), the protocol to 3.18.2, the types to 3.18.0, `vscode-jsonrpc` to 9.0.1 and `vscode-languageserver-textdocument` to 1.0.13.

Users see one change: the extension's **Solidity** output channel is now a log channel, so its verbosity can be filtered from the Output panel.

## Technical decisions

Three things had to changed with the versions:

- LSP 3.18 widens `Diagnostic.message` to `string | MarkupContent`. Five compiler diagnostics match against the message as a string. The first commit adds a `diagnosticMessage` helper that reads the text out of either shape — everything we act on comes from solc as plain text — and is behaviour preserving on its own. `TestLanguageClient` in the protocol tests reads the message the same way and gets the same narrowing.
- `LanguageClientOptions.outputChannel` now takes a `LogOutputChannel`, so the client builds its channel with `{ log: true }`. That is where the user-visible change comes from.
- Notification handlers may return a promise as of `vscode-jsonrpc@9`. `connection.onExit` returned the telemetry flush promise, but the type has changed.


server/src/server.ts
```diff
@@ -81,9 +81,9 @@ function attachLanguageServerLifeCycleHooks(
   connection.onInitialize(onInitialize(serverState));
   connection.onInitialized(onInitialized(serverState, workspaceFileRetriever));
 
-  connection.onExit(() => {
+  connection.onExit(async () => {
     serverState.logger.info("Server closing down");
-    return serverState.telemetry.close();
+    await serverState.telemetry.close();
   });
```
kanej, line 87: This is the second change. Hold on exit of the server until telemetry has closed.

kanej, line 87: The `onExit` signature changed so that a `Promise<boolean>` didn't work. So we `await` so that the anon function becomes `Promise<void>`.

PR #754, opened 2026-08-20
chore: bump changesets

kanej:
Update to the latest changeset. This required updating some of the config.

We will review all of this in more detail for the automated release work.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -10,7 +10,7 @@
     ]
   ],
   "linked": [],
-  "access": "restricted",
+  "access": "public",
   "baseBranch": "main",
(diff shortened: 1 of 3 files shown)


.changeset/config.json
```diff
@@ -2,10 +2,20 @@
...
   "linked": [],
   "access": "restricted",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": []
+  "ignore": [],
+  "privatePackages": {
+    "version": true,
+    "tag": false
+  }
```
kanej, line 20: This is for the extension, which is marked as private (we never publish to npm) but we do build and publish to the marketplace.

